PR #11, opened 2025-02-20
Rework partitioning scheme

stgraber:


jarrodu:
What was the issue? It looks like you got it to work.

stgraber:
@jarrodu it was a bit of a weird one. Basically systemd-repart was confusing the local-data and seed-data partitions as the installed system didn't have a repart.d file corresponding to the seed-data.

So I basically had to put a mostly blank file in repart.d for the seed-data partition and that made the local-data partition work too.

Bonus point is that the 16K size for the second verity partition worked fine, so I now have a partitioning scheme where the A and B sides have identically sized partitions.

PR #23, opened 2025-03-25
Add logic to install incus-osd to disk

gibmat:


incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,103 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/lxc/incus/v6/shared/subprocess"
+)
+
+// IsInstallNeeded checks for the presence of the file /boot/.incus-osd-installed to indicate
+// if we should attempt to install incus-osd to a local disk.
+func IsInstallNeeded() bool {
+	_, err := os.Stat("/boot/.incus-osd-installed")
```
gibmat: I also played with attempting to tag/create something in the partition table to indicate if `incus-osd` has been installed or not, but couldn't get anything reasonably simple to work.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimRight(output, "\n"), nil
+}
+
+// GetTargetDevice determines the underlying device to install incus-osd on.
+func GetTargetDevice(_ context.Context) (string, error) {
+	return "/dev/nvme0n1", nil // FIXME
```
gibmat: Need to add some basic logic to determine an install disk. Eventually we'll expand this logic to allow hints about what target device to select.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// DoInstall performs the steps to install incus-osd from the given target to the source device.
+func DoInstall(ctx context.Context, sourceDevice string, targetDevice string) error {
```
gibmat: Currently this will blindly jump in and destroy anything on the target device. Should probably add checks to confirm the device doesn't already contain a partition table.

incus-osd/cmd/incus-osd/main.go
```diff
@@ -46,6 +47,32 @@ func run() error {
...
+		}
+
+		slog.Info(fmt.Sprintf("Installing incus-osd from %s to %s", sourceDevice, targetDevice))
+
+		err = install.DoInstall(ctx, sourceDevice, targetDevice)
+		if err != nil {
+			return err
+		}
+
+		slog.Info("incus-osd was successfully installed. Please reboot the system and remove the external boot media.")
```
gibmat: We'll have limited ability to notify the user of any status, so it would be nice if we could figure out a robust way to reboot the system after install whether or not the install media is still present.

incus-osd/internal/install/install.go
stgraber: Can we just open the seed partition on the target and delete the `install.json`, `install.yaml` file from the seed?

incus-osd/internal/install/install.go
stgraber: Possibly, but also need a way to override that. When provisioning a server, existing stuff on disk shouldn't be a blocker.

incus-osd/cmd/incus-osd/main.go
stgraber: Can we resolve `/dev/disk/by-partlabel/usr` to its actual file (`/dev/sdaX`) and then have a loop which checks if that still exists, then as soon as it doesn't, we write `b` into /proc/sysrq-trigger?

incus-osd/internal/install/install.go
gibmat: Yeah, that works too :smile:

incus-osd/internal/install/install.go
gibmat: Added both a check that no existing partition table exists on the target device, along with a `ForceInstall` option for the install config to override that check.

incus-osd/cmd/incus-osd/main.go
gibmat: Added logic to automatically reboot if the install media is removed.

There will still be an issue with duplicate partition labels if the system is rebooted and the install media is still connected. `efibootmgr` can be used to properly select the installed disk upon reboot, but there's unhappiness with duplicate labels.

incus-osd/internal/install/install.go
stgraber: At this point, it should just be whatever other drive we find (so excluding the boot media). If we find more than one such drive, we fail with an error telling the user to specify it in the seed data.

incus-osd/cmd/incus-osd/main.go
stgraber: Yeah, the user needs to make sure not to plug it back in :)

Lennart was telling me on Monday that there's actually work being done to improve this situation with a new set of /dev symlinks which will be guaranteed to be tied to the booted image. So the conflicts can still happen in other place within /dev, but there will be a set of reliable ones that will work even if a conflict is present.

Anyway, that's being worked on and not in a stable release of systemd yet, so probably not going to make trixie meaning it's going to be something to look at quite a bit further down the line :)

incus-osd/internal/install/install.go
gibmat: Implemented logic to search for NVME, SCSI, and virtual disks (in that order). Also have basic substring matching against device IDs when selecting from more than one possible install target.

incus-osd/cmd/incus-osd/main.go
gibmat: If we're OK with depending on the user to remove the install disk once the install process is complete, I think this PR will be ready to merge.

incus-osd/internal/seed/install.go
```diff
@@ -0,0 +1,20 @@
+package seed
+
+// InstallConfig defines a struct to hold install configuration.
+type InstallConfig struct {
+	ForceInstall        bool   `json:"force_install"         yaml:"forceInstall"`
+	TargetDiskSubstring string `json:"target_disk_substring" yaml:"targetDiskSubstring"`
+}
```
stgraber: I think we can do a few minor tweaks on that.

Let's maybe go with:
```
force_reboot: false  # Reboot the system automatically rather than wait for install media to be unplugged.
force_install: false # Ignore existing data on the target.
target: # Optional selector for root disk, if not specified, we're expecting a single drive in the system or will fail
  id: xyz # Name as found in /dev/disk/by-id, glob supported
```

stgraber: @gibmat I think with that syntax, we'll have the `target` struct that we can evolve to support selecting based on type, size, ... down the line.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+	// Do a final sync and then reboot the system.
+	_, err := subprocess.RunCommandContext(ctx, "sync")
+	if err != nil {
+		return err
+	}
+
+	_, err = subprocess.RunCommandContext(ctx, "sh", "-c", "echo b > /proc/sysrq-trigger")
+	if err != nil {
+		return err
+	}
```
stgraber: Should definitely do that in native Go :)

unix.Sync() and io.WriteFile() should handle those just fine.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+	sourcePartitionPrefix := getPartitionPrefix(sourceDevice)
+	targetPartitionPrefix := getPartitionPrefix(targetDevice)
+
+	// Copy the partition contents.
+	for i := 1; i <= 8; i++ {
+		_, err = subprocess.RunCommandContext(ctx, "dd", fmt.Sprintf("if=%s%s%d", sourceDevice, sourcePartitionPrefix, i), fmt.Sprintf("of=%s%s%d", targetDevice, targetPartitionPrefix, i))
+		if err != nil {
+			return err
+		}
+	}
```
stgraber: We should do that natively, open source and target files + loop of io.CopyN

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+
+	// Remove the install configuration file, if present, from the target seed partition.
+	targetSeedPartition := fmt.Sprintf("%s%s2", targetDevice, targetPartitionPrefix)
+	for _, filename := range []string{"install.json", "install.yaml"} {
+		_, err = subprocess.RunCommandContext(ctx, "sh", "-c", fmt.Sprintf(`if tar -tf %s %s; then tar -f %s --delete %s; fi`, targetSeedPartition, filename, targetSeedPartition, filename))
+		if err != nil {
+			return err
+		}
+	}
```
stgraber: Does `tar -f foo.tar --delete bar` fail with a specific error code or something if not found?

If so, we could simplify that, avoid the whole spawning a shell subprocess and just run tar directly and check the return code instead.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf("a partition table already exists on device '%s', and `force_install` from install configuration isn't true", targetDevice)
+	}
+
+	// Turn off swap and unmount /boot.
+	_, err = subprocess.RunCommandContext(ctx, "swapoff", "-a")
+	if err != nil {
+		return err
+	}
+
+	_, err = subprocess.RunCommandContext(ctx, "umount", "/boot/")
```
stgraber: Should do that natively

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+	// Verify the target device doesn't already have a partition table, or that `ForceInstall` is set to true.
+	output, err := subprocess.RunCommandContext(ctx, "sgdisk", "-v", targetDevice)
+	if err != nil {
+		return err
+	}
+
+	if !strings.Contains(output, "Creating new GPT entries in memory") && !config.ForceInstall {
+		return fmt.Errorf("a partition table already exists on device '%s', and `force_install` from install configuration isn't true", targetDevice)
+	}
+
```
stgraber: Why are we even checking if `ForceInstall` is set to true?

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+	// If we have any empty install file, that should still trigger an install.
+	if errors.Is(err, io.EOF) {
+		return true
+	}
+
+	return err == nil
+}
+
+// GetSourceDevice determines the underlying device incus-osd is running on.
+func GetSourceDevice(ctx context.Context) (string, error) {
```
stgraber: I think we can do a faster native implementation of this.

Here is an example (missing error handling):
```go
	s := unix.Stat_t{}
	unix.Stat("/usr", &s)
	major := unix.Major(uint64(s.Dev))
	minor := unix.Minor(uint64(s.Dev))
	rootDev := fmt.Sprintf("%d:%d\n", major, minor)

	entries, _ := os.ReadDir("/sys/class/block")
	for _, entry := range entries {
		entryPath := filepath.Join("/sys/class/block", entry.Name())

		_, err := os.Stat(filepath.Join(entryPath, "partition"))
		if !errors.Is(err, os.ErrNotExist) {
			continue
		}

		dev, _ := os.ReadFile(filepath.Join(entryPath, "dev"))
		if string(dev) == rootDev {
			fmt.Printf("/dev/%s\n", entry.Name())
			return
		}

		subEntries, _ := os.ReadDir(entryPath)
		for _, subEntry := range subEntries {
			subEntryPath := filepath.Join("/sys/class/block", subEntry.Name())

			_, err := os.Stat(filepath.Join(subEntryPath, "partition"))
			if err != nil {
				continue
			}

			dev, _ := os.ReadFile(filepath.Join(subEntryPath, "dev"))
			if string(dev) == rootDev {
				fmt.Printf("/dev/%s\n", entry.Name())
				return
			}
		}
	}
```

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,234 @@
...
+	output, err := subprocess.RunCommandContext(ctx, "sh", "-c", `MAJOR_MINOR=$(dmsetup deps usr | sed -E 's/^.+: \(([0-9]+), ([0-9]+)\).+$/\1:\2/'); PARTITION=$(udevadm info -rq name /sys/dev/block/${MAJOR_MINOR}); lsblk --list --noheadings --paths --output PKNAME ${PARTITION} | head -n 1`)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimRight(output, "\n"), nil
+}
+
+// GetTargetDevice determines the underlying device to install incus-osd on.
+func GetTargetDevice(ctx context.Context, sourceDevice string) (string, error) {
```
stgraber: I thought we said that for now, we either want a single match through the provided pattern or a single device other than that we're booted from?

If so, we don't really need that logic. We can just iterate over /sys/class/block, ignore all the stuff that's not full drives, eliminate ourselves and if there's more than one hit, we fail.

incus-osd/cmd/incus-osd/main.go
```diff
@@ -46,6 +47,32 @@ func run() error {
...
+		if err != nil {
+			return err
+		}
+
+		targetDevice, err := install.GetTargetDevice(ctx, sourceDevice)
+		if err != nil {
+			return err
+		}
+
+		slog.Info(fmt.Sprintf("Installing incus-osd from %s to %s", sourceDevice, targetDevice))
```
stgraber: With slog, it's recommended to keep the string static and use variables instead, so in this case `slog.Info("Installing incus-osd", "source", sourceDevice, "target", targetDevice)`

incus-osd/internal/install/install.go
gibmat: There's not a distinct error code, but it's easy enough to look at the error message itself.

incus-osd/internal/install/install.go
gibmat: If no partition table already exists on a device, `gdisk` prints a line about creating entries in memory. If that line isn't in the output that means there's a partition table on the device, and we won't proceed unless `ForceInstall` has been set to true in the seed configuration.

incus-osd/internal/install/install.go
gibmat: For the moment, yes, we're just expecting one other drive to install to. We do have basic logic in place to make a selection based on the device's ID; if that's not specified we'll fail if we don't find exactly two devices.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	potentialTargets = append(potentialTargets, virtualTargets.Blockdevices...)
+
+	// Ensure we found at least two devices (the install device and potential install device(s)). If no Target
+	// configuration was found, only proceed if exactly two devices were found.
+	if len(potentialTargets) < 2 {
+		return "", errors.New("no potential install devices found")
+	} else if i.config.Target == nil && len(potentialTargets) != 2 {
+		return "", errors.New("no Target configuration provided, and didn't find exactly one install device")
```
stgraber: Nit `target`

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,328 @@
...
+		for {
+			_, err := io.CopyN(targetPartition, sourcePartition, 1024*1024)
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					break
+				}
+
+				return err
+			}
+		}
```
stgraber: Content of the loop should be moved to an in-line function and then be called from within the loop, that way the defers will work and you can drop the nolint.

incus-osd/internal/install/install.go
```diff
@@ -0,0 +1,328 @@
...
+	slog.Info("incus-osd was successfully installed. Please reboot the system and remove the external boot media.")
+
+	return i.rebootUponDeviceRemoval(ctx, sourceDevice)
+}
+
+// getSourceDevice determines the underlying device incus-osd is running on.
+func (*Install) getSourceDevice() (string, error) {
+	// Start by determining the underlying mapped device that /usr is on.
+	s := unix.Stat_t{}
+	err := unix.Stat("/usr", &s)
```
stgraber: Probably easier to look at /boot/efi which should be an actual partition instead of a dm device.

stgraber (approved):

PR #27, opened 2025-04-03
Adjust test VM configuration

gibmat:
Modifies the `test` target to setup the VM with Incus OS in a similar manner as if it was running from a removable device.


Makefile
```diff
@@ -33,21 +33,37 @@ build: incus-osd
...
 		-c limits.cpu=4 \
 		-c limits.memory=8GiB \
 		-d root,size=50GiB
 	incus config device add test-incus-os vtpm tpm
+	incus config device add test-incus-os boot-media disk source=$$(pwd)/mkosi.output/IncusOS_boot_media.img boot.priority=10
+
 	incus start test-incus-os --console
 	sleep 3
-	incus console test-incus-os
+	incus console test-incus-os || (sleep 3 && incus console test-incus-os)
```
gibmat, line 53: Occasionally I'm seeing the attempt to get a console fail with "Error: Failed to swap console ring buffer with socket: listen unix /run/incus/test-incus-os/qemu.console: bind: address already in use"

Makefile
```diff
@@ -33,21 +33,37 @@ build: incus-osd
...
+
+	incus config device remove test-incus-os boot-media
+	sleep 3
+	incus stop -f test-incus-os
+
+	incus start test-incus-os
+	sleep 3
+	incus console test-incus-os || (sleep 3 && incus console test-incus-os)
```
gibmat, line 60: And this attempt at getting a console almost always fails

Makefile
stgraber: Yeah, there's something weird going on with the switch between ring buffer and live console, but I've not managed to reproduce it reliably enough to get to the bottom of it.

stgraber (approved):

PR #34, opened 2025-04-08
Add hostname and domain configuration to the network configuration

### LINKED ISSUE #32: Add hostname and domain configuration to the network configuration
We should add top-level hostname and domain configuration keys to the network configuration which will then be used to configure the system hostname and provide us with the expected FQDN.

This will be quite useful as part of #30 as otherwise we'll just be reporting localhost.localdomain back to the switch :)

gibmat:
Closes #32

### DIFF
--- a/incus-osd/internal/seed/network_types.go
+++ b/incus-osd/internal/seed/network_types.go
@@ -4,6 +4,9 @@ package seed
 type NetworkConfig struct {
 	Version string `json:"version" yaml:"version"`
 
+	Hostname string `json:"hostname" yaml:"hostname"`
+	Domain   string `json:"domain"   yaml:"domain"`
+
 	Interfaces []NetworkInterface `json:"interfaces" yaml:"interfaces"`
 	Bonds      []NetworkBond      `json:"bonds"      yaml:"bonds"`
 	Vlans      []NetworkVlan      `json:"vlans"      yaml:"vlans"`
--- a/incus-osd/internal/systemd/networkd.go
+++ b/incus-osd/internal/systemd/networkd.go
@@ -68,6 +68,20 @@ func ApplyNetworkConfiguration(ctx context.Context, networkCfg *seed.NetworkConf
 		return errors.New("no network configuration provided")
 	}
 
+	// Set hostname from network config, if defined.
+	if networkCfg.Hostname != "" {
+		hostname := networkCfg.Hostname
+		if networkCfg.Domain != "" {
+			hostname += "." + networkCfg.Domain
+		}
+
+		// Apply the configured hostname.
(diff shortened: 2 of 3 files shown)


incus-osd/internal/systemd/networkd.go
```diff
@@ -75,7 +76,28 @@ UseMTU=true`), 0o644)
...
-	err := generateNetworkConfiguration(ctx, networkCfg)
+	// Return immediately if given a nil network configuration.
+	if networkCfg == nil {
+		return nil
+	}
+
+	if networkCfg.Hostname == "" {
+		return errors.New("hostname must be provided in the network configuration")
+	}
+
```
stgraber: I think we need to tolerate an empty string, at which point we'd just leave things as they are, meaning `localhost.localdomain`.

incus-osd/internal/seed/network_types.go
```diff
@@ -2,7 +2,9 @@ package seed
 
 // NetworkConfig defines a struct to hold the three types of supported network configuration.
 type NetworkConfig struct {
-	Version string `json:"version" yaml:"version"`
+	Version  string `json:"version"  yaml:"version"`
```
stgraber: Tiny nit but I've left the `Version` field on its own block at the beginning of the structs so that's easier to tell if we forgot it on some of them.

stgraber: I'll do that small tweak for you now since I already have this branch pulled locally.

stgraber (approved):

stgraber (approved):

stgraber:
@gibmat something's weird, we keep failing the tests with this PR at basically the same spot, so looks like it's causing a regression.

incus-osd/internal/systemd/networkd.go
```diff
@@ -33,8 +33,8 @@ func generateNetworkConfiguration(_ context.Context, networkCfg *seed.NetworkCon
 		return err
 	}
 
-	// When no configuration is provided, just DHCP on everything..
-	if networkCfg == nil {
+	// When no configuration is provided, just DHCP on everything.
+	if networkCfg == nil || len(networkCfg.Interfaces) == 0 {
```
stgraber: That feels wrong. If I provided an empty config on purpose, I'd expect nothing to happen.

Also the way I'm reading this, if I push a config with no interfaces but with a bond defined, I'd end up with the default config, which definitely feels wrong.

gibmat: I've cleaned this up in #35, and will rebase once that's merged.

stgraber:
@gibmat ready for rebase

PR #35, opened 2025-04-08
Save network seed to state

### LINKED ISSUE #29: Network seed data needs to be copied into state
On first boot we should look at the network seed data (or our default fallback), but the result needs to be added to the state struct and be read back from there from that point on.

gibmat:
Closes #29, #40


stgraber:
@gibmat I think that as part of this we need to change how we handle the default config.

So instead of having a fallback hardcoded systemd-network config for when no network seed was provided.
We should instead have a function which returns a network struct to be used in that situation and have that be stored in the seed.

That will become more important when we add the API as we don't ever want a nil value for the network config.


So I think what we should basically do is if a network seed is absent, fill in a network struct with every detected interface (net.Interfaces()) getting a dhcp4+dhcp6 or equivalent network config.

incus-osd/internal/systemd/networkd.go
```diff
@@ -90,6 +90,10 @@ func ApplyNetworkConfiguration(ctx context.Context, networkCfg *seed.NetworkConf
 		return networkCfg, err
 	}
 
+	// At system start there's a small race between udev being fully started and
+	// our reconfiguring of the network. Sleep for a couple seconds before triggering udev.
+	time.Sleep(2 * time.Second)
```
gibmat, line 78: Without this there's a race between `systemd-udevd` starting and us attempting to trigger the udev update. I haven't been able to figure out a reliable way to determine when udev is fully "ready" (tried depending on `systemd-udevd-settle.service`  as well as looking for something under `/run/udev/`).

stgraber: What happens if you call `udevadm settle` as a way to wait for udevd to be ready?

incus-osd/cmd/incus-osd/main.go
```diff
@@ -152,13 +152,16 @@ func startup(ctx context.Context, s *state.State) error {
...
+			return err
+		}
 	}
 
-	err = systemd.ApplyNetworkConfiguration(ctx, network)
+	// Perform network configuration.
+	s.NetworkConfig, err = systemd.ApplyNetworkConfiguration(ctx, s.NetworkConfig)
 	if err != nil {
 		return err
 	}
```
stgraber: This looks pretty weird. Wouldn't it be cleaner to just have the code above do something like:
```
if seed.IsMissing(err) {
    s.NetworkConfig = systemd.DefaultNetworkConfiguration()
}
```

gibmat: Did some refactoring and moved that logic directly to the seed code.

incus-osd/internal/systemd/networkd.go
gibmat: That doesn't seem to help, and is essentially what `systemd-udevd-settle.service` does.

stgraber (approved):

PR #47, opened 2025-04-16
Rework TUI state

### LINKED ISSUE #38: Rework use of state within tui
There should only ever be a single `state` struct that we pass around to the other leaf packages during initialization. That `state` struct should include both the persistent on-disk state that's saved to the .json file as well as additional internal struct members.

Currently I noticed two problems:
 - We're using a tui.GetUI function from the install package to fetch the UI directly rather than having the UI instance be stored as a pointer on the state struct and have the state struct be passed to the `install` package during its instantiation.
 - We're directly parsing the on-disk state again within the tui package to render the system state. We should instead have passed a state struct pointer during initialization of the tui instance and so be able to access that at any point we want.

Instead I'd expect the `main` package to handle all of the instantiation bits, which includes:
 - logger
 - tui
 - state

And then pass those specific instances to other packages that need it during their own initialization.

So in this case, I think we should drop `tui.GetTUI` and instead export `tui.NewTUI` and have it take `*state.State` as an argument.

Then when calling `install.NewInstall` we can pass it the `tui` instance so it can push its update through it. `NewInstall` will just put that `tui` reference in its struct which will then be readily accessible.

gibmat:
Closes #38

We can't directly place a reference to the TUI package into the state struct, since that introduces an import loop. I've done some refactoring in main to pass the state when creating the TUI, then passing the TUI into the installer code.

### DIFF
--- a/incus-osd/cmd/incus-osd/main.go
+++ b/incus-osd/cmd/incus-osd/main.go
@@ -28,14 +28,37 @@ var (
 )
 
 func main() {
+	ctx := context.Background()
+
 	// Check privileges.
 	if os.Getuid() != 0 {
 		_, _ = fmt.Fprintf(os.Stderr, "incus-osd must be run as root")
 		os.Exit(1)
 	}
 
+	// Create runtime path if missing.
+	err := os.Mkdir(runPath, 0o700)
+	if err != nil && !os.IsExist(err) {
+		_, _ = fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+
+	// Create storage path if missing.
+	err = os.Mkdir(varPath, 0o700)
+	if err != nil && !os.IsExist(err) {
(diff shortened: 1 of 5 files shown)


incus-osd/cmd/incus-osd/main.go
```diff
@@ -34,8 +34,31 @@ func main() {
...
+	}
+
+	// Create storage path if missing.
+	err = os.Mkdir(varPath, 0o700)
+	if err != nil && !os.IsExist(err) {
+		_, _ = fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		os.Exit(1)
+	}
+
+	ctx := context.TODO()
```
stgraber: I'd move that to the first line of the function and make it a `context.Background()` at this point. We're a daemon so not expecting to receive another context.

stgraber (approved):

PR #48, opened 2025-04-16
Enable AppArmor

### LINKED ISSUE #44: Enable AppArmor
It would be good to have AppArmor correctly enabled for extra security.

To do this we'll need to:
 - Have it included in the base (instead of only in incus)
 - Copy /etc/apparmor and /etc/apparmor.d over to somewhere in /usr
 - Set up a systemd-tmpfiles rule to symlink the expected /etc paths to the copy in /usr
 - Remove apparmor=0 from the kernel command line

gibmat:
Closes #44


stgraber:
Nice and easy!
Did you confirm that Incus works properly with AppArmor enabled?

gibmat:
> Did you confirm that Incus works properly with AppArmor enabled?

Yup, I tested creating a simple container and then installing a package via apt. On the IncusOS side `aa-status` showed the generated profiles correctly in enforcing mode.

stgraber:
@gibmat can you give this one a quick rebase so it picks up the test fix that was merged as part of the TUI PR?

mkosi.images/incus/install.sh
```diff
@@ -12,7 +12,7 @@ cat <<EOF > /etc/apt/sources.list.d/zabbly-incus-daily.sources
 Enabled: yes
 Types: deb
 URIs: https://pkgs.zabbly.com/incus/daily
```
stgraber: That's weird, I could have sworn I had done that change already... I guess I never pushed that commit.

Can you also swap us over to `incus/stable` instead of `incus/daily`?

There's no reason to be running daily anymore.

stgraber (approved):

PR #57, opened 2025-04-17
Add encryption recovery API

### LINKED ISSUE #52: Implement encryption recovery mechanism
Our main data partition is encrypted based on TPM PCR register states.
In the event that those don't line up, a recovery key must be provided.

I believe that we currently have an empty recovery key (empty string) enrolled which is far from ideal :)

During initialization we should therefore:
 - Generate a new random recovery key. Microsoft uses a 48 digit key split across 8 blocks of 6 digits. I'd suggest we use the same format so that it can easily fit in existing key management infrastructure.
 - The recovery key should be stored in the state along with a flag to tell us whether the user has retrieved it yet
 - The TUI should show a persistent warning message until such time as the user has retrieved the recovery key from the API

At the API level, we should add an endpoint at `/1.0/system/encryption` with a basic struct like:
```
type Encryption struct {
    RecoveryKeys []string
}
```

We may extend that logic later to support publishing the recovery key to an external key escrow server. We can allow editing of the list of recovery keys and will update the LUKS device when that happens.

### LINKED ISSUE #53: Create encrypted ZFS pool on the local-data partition
We have an extra local partition labeled `local-data` which is always created with whatever space is left on the system during initialization.

The expectation is that this would be consumed by Incus as a basic local storage pool.

We should automatically initialize it during system initialization by:
 - Generate a random encryption keyfile and store it in /var/lib/incus-os (`zpool.local.key`)
 - Creating a new `local` ZFS pool on it with ZFS encryption enabled using the encryption key
 - Having incus-osd unlock the zpool on boot (`zfs load-key local`) prior to starting the applications

This will provide us with a convenient ZFS pool that can be used by Incus and which is effectively as safe as the main OS storage as its encryption key will be stored on the encrypted root disk.

gibmat:
Closes #52

We no longer set an empty encryption password in the repart configuration for the root partition.

The tpm binding defaults to looking at just PCR 7; do we want to modify that?

`systemd-cryptenroll` has a nice built-in option to generate recovery keys that look similar to BitLocker's: `fikibljt-rikgbifh-uijvcivb-eljkfufv-uugitiin-ujcdugdt-vibnnrev-iivubuiv`.

Adds a REST endpoint to get a list of current recovery passwords, as well as adding and removing individual passwords.

image

### DIFF
--- a/incus-osd/internal/rest/api_encryption.go
+++ b/incus-osd/internal/rest/api_encryption.go
@@ -1,0 +1,55 @@
+package rest
+
+import (
+	"errors"
+	"io"
+	"net/http"
+
+	"github.com/lxc/incus-os/incus-osd/internal/rest/response"
+	"github.com/lxc/incus-os/incus-osd/internal/systemd"
+)
+
+func (s *Server) apiSystemEncryption(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	switch r.Method {
+	case http.MethodGet:
+		// Mark that the keys have been retrieved via the API.
+		s.state.System.Encryption.RecoveryKeysRetrieved = true
+
+		// Return the current system encryption state.
+		_ = response.SyncResponse(true, s.state.System.Encryption).Render(w)
+	case http.MethodPut, http.MethodDelete:
+		// Add or remove an encryption key.
+		if r.ContentLength <= 0 {
+			_ = response.BadRequest(errors.New("no encryption key provided")).Render(w)
+
+			return
+		}
+
+		b, err := io.ReadAll(r.Body)
+		if err != nil {
+			_ = response.BadRequest(err).Render(w)
+
+			return
+		}
+
+		if r.Method == http.MethodPut {
+			err = systemd.AddEncryptionKey(r.Context(), s.state, string(b))
+		} else {
+			err = systemd.DeleteEncryptionKey(r.Context(), s.state, string(b))
+		}
+		if err != nil {
+			_ = response.BadRequest(err).Render(w)
+
+			return
+		}
+
+		_ = response.EmptySyncResponse.Render(w)
+	default:
+		// If none of the supported methods, return NotImplemented.
+		_ = response.NotImplemented(nil).Render(w)
+	}
+
+	_ = s.state.Save(r.Context())
+}
--- a/incus-osd/internal/state/struct.go
+++ b/incus-osd/internal/state/struct.go
@@ -27,6 +27,7 @@ type State struct {
 	} `json:"services"`
 
 	System struct {
-		Network *api.SystemNetwork `json:"network"`
+		Encryption *api.SystemEncryption `json:"encryption"`
+		Network    *api.SystemNetwork    `json:"network"`
 	} `json:"system"`
 }
--- a/incus-osd/internal/tui/tui.go
+++ b/incus-osd/internal/tui/tui.go
@@ -187,6 +187,10 @@ func (t *TUI) redrawScreen() {
 		t.frame.AddText(line, false, tview.AlignLeft, tcell.ColorWhite)
 	}
 
+	if !t.state.System.Encryption.RecoveryKeysRetrieved {
+		t.frame.AddText("WARNING: Encryption recovery key has not been retrieved yet!", false, tview.AlignLeft, tcell.ColorRed)
+	}
+
 	if t.textView != nil {
 		t.frame.SetPrimitive(t.textView)
 	}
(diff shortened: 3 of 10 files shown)


incus-osd/internal/rest/api_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+	case http.MethodGet:
+		// Return the current recovery keys.
+		ret := Encryption{
+			RecoveryKeys: s.state.RecoveryKeys,
+		}
+		_ = response.SyncResponse(true, ret).Render(w)
+
+		// Mark that the keys have been retrieved via the API.
+		s.state.RecoveryKeysRetrieved = true
+	case http.MethodPut, http.MethodDelete:
```
gibmat, line 22: I'm sometimes getting back an empty server response, although the underlying add/delete are succeeding.

incus-osd/internal/state/struct.go
```diff
@@ -18,4 +18,7 @@ type State struct {
 	RunningRelease string                 `json:"running_release"`
 
 	NetworkConfig *seed.NetworkConfig `json:"network_config"`
+
+	RecoveryKeys          []string `json:"recovery_keys"`
+	RecoveryKeysRetrieved bool     `json:"recovery_keys_retrieved"`
```
stgraber: Once my branch is merged, we'll have network as `System.Network`, so this should probably be a `System.Encryption` field pointing to a matching `api.SystemEncryption` struct.

incus-osd/internal/encryption/encryption.go
```diff
@@ -0,0 +1,78 @@
+package encryption
```
stgraber: Those are all systemd commands, so would probably make more sense in the `systemd` package then.

incus-osd/internal/rest/api_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/lxc/incus-os/incus-osd/internal/encryption"
+	"github.com/lxc/incus-os/incus-osd/internal/rest/response"
+)
+
+// Encryption defines an API-level representation of configured encryption keys.
+type Encryption struct {
+	RecoveryKeys []string `json:"recovery_keys"`
+}
+
```
stgraber: This should go into the `api` package once my branch is merged.

incus-osd/internal/rest/api_encryption.go
```diff
@@ -0,0 +1,63 @@
...
+	switch r.Method {
+	case http.MethodGet:
+		// Return the current recovery keys.
+		ret := Encryption{
+			RecoveryKeys: s.state.RecoveryKeys,
+		}
+		_ = response.SyncResponse(true, ret).Render(w)
+
+		// Mark that the keys have been retrieved via the API.
+		s.state.RecoveryKeysRetrieved = true
```
stgraber, line 29: Please call `s.state.Write()` to make sure it's written to disk.

incus-osd/internal/rest/api_encryption.go
stgraber: That's weird, I don't see anything obviously wrong in the logic here.
Anything interesting in the journal?

incus-osd/internal/tui/tui.go
```diff
@@ -187,6 +187,10 @@ func (t *TUI) redrawScreen() {
 		t.frame.AddText(line, false, tview.AlignLeft, tcell.ColorWhite)
 	}
 
+	if !t.state.System.Encryption.RecoveryKeysRetrieved {
+		t.frame.AddText("WARNING: Encryption recovery key has not yet been retrieved via API!", false, tview.AlignLeft, tcell.ColorRed)
```
stgraber: I'd make this "has not been retrieved yet!".

Whether it's done by the user using the API or by Operations Center or through our upcoming CLI isn't really relevant.

stgraber (changes requested):

incus-osd/internal/rest/api_encryption.go
gibmat: It's hitting the 10 second default write timeout that's set when the HTTP server is created. Usually the first time adding a new password takes longer and hits the timeout, but then subsequent additions will take less than 10 seconds. A pathological case is adding several passwords, then deleting one, since we need to re-add all the remaining ones which can easily take over 30 seconds.

The context itself isn't canceled, so the process(es) finish running, and then instead of sending back the response Go just closes the connection.

stgraber: I'd be fine getting rid of those timeouts. They don't make too much sense on a restricted API run over a local unix socket.

stgraber (approved):

incus-osd/internal/rest/api_encryption.go
gibmat: #62

PR #59, opened 2025-04-18
Implement shutdown/reboot handling

stgraber:


stgraber:
This one became a bit of a requirement with the NVME and iSCSI stuff as without it, we don't get a properly ordered shutdown/reboot causing everything to hang on the networked storage.

PR #61, opened 2025-04-18
Add more network config options

### LINKED ISSUE #49: Add more network config options
We should Introduce a `NetworkDNS` struct which includes:
 - Hostname string
 - Domain string
 - SearchDomains []string
 - Nameservers []string

That way environments with static nameservers can have it provided that way.

We should also introduce a `NetworkNTP` struct which includes:
 - Timeservers []string

And to round things up, also a `NetworkProxy` which includes:
 - HTTPProxy
 - HTTPSProxy
 - NoProxy

We should also do a quick sanity check that a VM booted with configured local DNS and NTP servers does not establish any connections outside of those networks. We want to ensure that Incus OS can run in a fully isolated environment and doesn't leak information to the outside world nor depends on external resources.

gibmat:
Closes #49

Still need to verify that we don't have any other unexpected network connections being made. (`MulticastDNS` is already disabled in the base image.)


stgraber (approved):

stgraber:
@gibmat just thought of something. On the proxy front, we need to also have our own code use that proxy information.

For now I think this is limited to the Github calls from within `provider_github.go`.
So we'd need to make sure that the provider package is passed in the state and that we set up an http client that uses the proxy info from the state.

gibmat:
> @gibmat just thought of something. On the proxy front, we need to also have our own code use that proxy information.
> 
> For now I think this is limited to the Github calls from within `provider_github.go`. So we'd need to make sure that the provider package is passed in the state and that we set up an http client that uses the proxy info from the state.

I do add any proxy environment variables to the daemon's environment via `os.Setenv()`; I'd have to double-check, but I hope the http client would pickup/respect that automagically. If not, I can pass in the state like you mentioned.

stgraber:
> I do add any proxy environment variables to the daemon's environment via `os.Setenv()`; I'd have to double-check, but I hope the http client would pickup/respect that automagically. If not, I can pass in the state like you mentioned.

Yeah, will need checking. My experience with os.Getenv/os.Setenv hasn't been exactly great :)

mkosi.images/base/mkosi.extra/usr/lib/systemd/timesyncd.conf.d/00-disable-fallback-ntp.conf
```diff
@@ -0,0 +1,2 @@
+[Time]
+FallbackNTP=
```
gibmat: For some reason, even with no fallback NTP servers defined and a local one specified in the network yaml, I'm still seeing a DNS lookup request for `0.debian.pool.ntp.org` when systemd-timesyncd starts up. The only place that string exists in the entire IncusOS image is the systemd-timesyncd binary itself.

stgraber: What do you get when you run `timedatectl show-timesync`?

gibmat:
```
LinkNTPServers=10.234.136.1
FallbackNTPServers=0.debian.pool.ntp.org 1.debian.pool.ntp.org 2.debian.pool.ntp.org 3.debian.pool.ntp.org
ServerName=10.234.136.1
ServerAddress=10.234.136.1
RootDistanceMaxUSec=5s
PollIntervalMinUSec=32s
PollIntervalMaxUSec=34min 8s
PollIntervalUSec=1min 4s
Frequency=0
```

stgraber: Okay, so setting the fallback to empty didn't work apparently.
What happens if you set it to the same value as the primary server?

gibmat: Setting an actual hostname for `FallbackNTP=` does override the defaults (I had tried a dummy value to confirm that the DNS request changed). Also, for fun I tried setting `FallbackNTP=""`, which also appears to disable the default fallback servers. The configuration looks a little weird, but I don't see any errors/warnings from systemd-timesyncd:

```
bash-5.2# timedatectl show-timesync
LinkNTPServers=10.234.136.1
FallbackNTPServers="\"\""
...
```

stgraber: Okay, probably would look cleaner to set the fallback to the same as the primary then

gibmat:
> > I do add any proxy environment variables to the daemon's environment via `os.Setenv()`; I'd have to double-check, but I hope the http client would pickup/respect that automagically. If not, I can pass in the state like you mentioned.
> 
> Yeah, will need checking. My experience with os.Getenv/os.Setenv hasn't been exactly great :)

I've tested that the proxy environment variables seem to be working properly. With a local squid proxy, I can configure `http_proxy` and `https_proxy` in the seed configuration to point to it; when squid is running `incus launch ....` works, and with squid stopped I get a proxyconnect error. :slightly_smiling_face:

stgraber:
Looks good for initial configuration but I'm not sure about the handling on later changes. 

We likely need to restart timesyncd, clear the proxy env variables, ... to properly apply a configuration change.

gibmat:
> We likely need to restart timesyncd, clear the proxy env variables, ... to properly apply a configuration change.

Updated logic to properly handle changes to any of the network configuration.

stgraber (approved):

PR #68, opened 2025-04-24
* README.md - added debian build environment setup instructions

caseykelso:
* .gitignore - added mkosi.csr and *.swp
Signed-off-by: Casey Kelso <casey@mutex.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,34 +37,55 @@ The instructions below assume a functional Incus environment with VM support.
 Two scripts are available to test Incus OS using the publicly published releases.
 
 Creating a new Incus OS VM can be done with:
-
+```bash
     ./scripts/spawn-image VERSION NAME
+```
 
 This will retrieve the relevant image from Github and create a VM using it.
 It will also automatically load the relevant packages (`incus` and `debug`).
 
 To update it to a newer version, you can do:
-
+```bash
     ./scripts/update-image VERSION NAME
+```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,34 +37,55 @@ The instructions below assume a functional Incus environment with VM support.
...
 This will retrieve the update files and apply the update.
 
+## Setup Debian 13 build environment ##
+```bash
+sudo apt update -y && sudo apt install git mkosi build-essential wget
+```
+
+## Install GO ##
+```bash
+mkdir -p ~/Downloads && pushd ~/Downloads && wget -c https://go.dev/dl/go1.24.2.linux-amd64.tar.gz
```
gibmat, line 61: If you're specifically targeting a Debian 13 build environment, it would be better to install go via apt: `sudo apt install golang-go`. Yes, you won't get the latest future go versions after trixie is released, but I don't think that will really matter for this project.

stgraber: I think I'd prefer we keep the instructions generic, so basically change this commit to provide instructions that should work on the majority of Debian and Ubuntu systems, by:
 - Downloading Go from upstream (as we only support the last two stable releases, what's in the distro quickly stops working)
 - Downloading our preferred mkosi tag using `pipx`

The rest shouldn't matter as the recent `mkosi` will build itself a build chroot to avoid depending on the system's packages.

PR #69, opened 2025-04-24
Background OS and app update check

### LINKED ISSUE #50: Implement background checking for OS and application updates
We should spawn a go-routine that will check for new OS and application updates.

As a first pass, we can add logic to:
 - Do a check every 6 hours
 - If an application update is found, download it and apply it, including service restart
 - If an OS update is found, download it and stage it, but don't reboot
 - Show a "restart required" or similar message in the TUI

Once we have that, we can start working on `/1.0/system/update` to handle configuration of update policies, but that'll be for a bit later.

gibmat:
Closes #50


incus-osd/cmd/incus-osd/main.go
```diff
@@ -287,11 +287,8 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error {
...
 
 	if p != nil {
-		// Run update function if we have a working provider.
-		err = update(ctx, s, t, p)
-		if err != nil {
-			return err
-		}
+		// Run periodic update checks if we have a working provider.
+		go updateChecker(ctx, s, t, p)
 	}
```
stgraber: We still need to do a first blocking run on start, otherwise we're going to be going through services and applications startup before we've even had a chance to install the applications.

PR #77, opened 2025-04-28
Make Incus OS work when booted from read-only media

### LINKED ISSUE #66: Make Incus OS work when booted from read-only media
We need to support Incus OS booting from a read-only media like a virtual cdrom drive.
This will likely involve an early unit which can override systemd-repart and give us a tmpfs backed temporary / so we can boot and install from that.

gibmat:
Closes #66

This adds the ability to run Incus OS from read-only media. The expected use case is to run the installer, then reboot into the freshly-installed system.

The default image built by running `make` can be booted from a read-only drive/USB stick. An ISO-compatible image can be created by running `make build-iso`. Due to differences in sector sizes, it's not possible to swap one image for the other; trying to get the necessary back magic to support a unified image seems like too much work for right now.


stgraber:
Hmm, verity-sig is 16k here:
```

Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048         4196351   2.0 GiB     EF00  esp
   2         4196352         4401151   100.0 MiB   8300  seed-data
   3         4401152         4401183   16.0 KiB    8385  IncusOS_20250426050...
   4         4401184         4605983   100.0 MiB   8319  IncusOS_20250426050...
   5         4605984         6703135   1024.0 MiB  8314  IncusOS_202504260502
   6         6703136         6703167   16.0 KiB    8385  _empty
   7         6703168         6907967   100.0 MiB   8319  _empty
   8         6907968         9005119   1024.0 MiB  8314  _empty
   9         9005120        17393727   4.0 GiB     8200  swap
  10        17393728        69822527   25.0 GiB    8304  root-x86-64
  11        69822528       104857559   16.7 GiB    8300  local-data
```

mkosi.images/base/mkosi.extra/usr/lib/repart.d/20-usr-verity-sig.conf
```diff
@@ -1,5 +1,5 @@
 [Partition]
 Type=usr-verity-sig
 Label=_empty
-SizeMinBytes=16K
-SizeMaxBytes=16K
+SizeMinBytes=1M
```
gibmat: There's still a little bit of a difference in the final post-install GPT table between an install running from read-write media (first) and one from read-only media (second). Not sure if there's really anything to be concerned about or not.

```
Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048         4196351   2.0 GiB     EF00  esp
   2         4196352         4401151   100.0 MiB   8300  seed-data
   3         4401152         4403199   1024.0 KiB  8385  IncusOS_20250430172...
   4         4403200         4607999   100.0 MiB   8319  IncusOS_20250430172...
   5         4608000         6705151   1024.0 MiB  8314  IncusOS_202504301728
   6         6705152         6707199   1024.0 KiB  8385  _empty
   7         6707200         6911999   100.0 MiB   8319  _empty
   8         6912000         9009151   1024.0 MiB  8314  _empty
   9         9009152        17397759   4.0 GiB     8200  swap
  10        17397760        69826559   25.0 GiB    8304  root-x86-64
  11        69826560       104857559   16.7 GiB    8300  local-data
```

```
Number  Start (sector)    End (sector)  Size       Code  Name
   1            2048         4196351   2.0 GiB     EF00  esp
   2         4196352         4401151   100.0 MiB   8300  seed-data
   3         4401152         4403199   1024.0 KiB  8385  IncusOS_20250430172...
   4         4403200         4607999   100.0 MiB   8319  IncusOS_20250430172...
   5         4608000         6705151   1024.0 MiB  8314  IncusOS_202504301728
   6              40            2047   1004.0 KiB  8385  _empty
   7         6705152         6909951   100.0 MiB   8319  _empty
   8         6909952         9007103   1024.0 MiB  8314  _empty
   9         9007104        17395711   4.0 GiB     8200  swap
  10        17395712        69824511   25.0 GiB    8304  root-x86-64
  11        69824512       104857559   16.7 GiB    8300  local-data
```

stgraber:
There was actually a bug in repart a while back that would require a 1MB minimum, but the issue was resolved which is when I moved both partitions to 16kB.

The first partition is always 16kB as generated by mkosi and so having the second partition be 1MB felt wrong.

mkosi.images/base/mkosi.extra/usr/lib/repart.d/20-usr-verity-sig.conf
stgraber: Hmm, I don't like that we're putting data on disk prior to the ESP partition in the second case.

gibmat: Indeed, it looks like the version of `mkosi` from trixie has that issue with creating a 16KiB partition.

stgraber: @gibmat yeah, you really want to use the version we're using on Github, that's a tag that was made after validating everything we needed :)

stgraber: Though they're both 25.3 so that's a bit confusing

gibmat: > Hmm, I don't like that we're putting data on disk prior to the ESP partition in the second case.

I looked at the options for `repart.d` and didn't see any obvious way to tell systemd-repart to create the partition after the existing ones.

stgraber: Yeah, I did the same...

Given this only happens in the install case, I wonder if we can't cheat a little by having the installer pre-create the 3 OS-B partitions directly following copy to the target disk.

The other two partitions can be handled by repart on boot just fine, they're large enough that they can't fit in that gap.

stgraber:
@gibmat I'd rather not include binary files if I can help it.
Can you generate the `.deb` prior to image building? Basically the same as we do for `incus-osd`.

This will require a Debian/Ubuntu system to build the image, but I'm fine with that, the main target for image building is Github Actions anyways.

mkosi.images/base/mkosi.extra/usr/lib/repart.d/20-usr-verity-sig.conf
gibmat: Sure, that was easy enough to add.

stgraber (approved):

PR #89, opened 2025-05-05
Network updates

### LINKED ISSUE #86: TUI should only show configured system interfaces
We shouldn't be showing other interfaces like those created by Incus.
It would also be useful to indicate what interface an address is on.

gibmat:
*  When bringing up the network, if there's no network seed consider things "up" if at least one interface is routable
* Only show IP addresses for configured interfaces (Closes #86)

### DIFF
--- a/incus-osd/internal/systemd/networkd.go
+++ b/incus-osd/internal/systemd/networkd.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lxc/incus/v6/shared/subprocess"
 
 	"github.com/lxc/incus-os/incus-osd/api"
+	"github.com/lxc/incus-os/incus-osd/internal/seed"
 )
 
 // networkdConfigFile represents a given filename and its contents.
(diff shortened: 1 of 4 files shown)


incus-osd/internal/systemd/networkd.go
```diff
@@ -142,12 +143,12 @@ func ApplyNetworkConfiguration(ctx context.Context, networkCfg *api.SystemNetwor
 	}
 
 	// Wait for the network to apply.
-	return waitForNetworkRoutable(ctx, networkCfg, timeout)
+	return waitForNetworkRoutable(ctx, networkCfg, timeout, seed.NetworkSeedExists())
```
stgraber, line 146: What if the user later on clears the network config through the API?
In that situation they'd have a network seed with some fixed networks but a current network config which uses the default config, leading to completely broken incus-osd.

In general, I think we may want to handle this through a timeout mechanism.
Try to wait up to 30s for all configured networks to be online and functional.

If we don't by the time we reach the timeout, check if we have a default IPv4 OR IPv6 route, if we have either of those, keep going as if we're online (after logging a suitable warning).

If we don't have any default routes on the system, then fail with a modal failure message (system requirements style).

PR #99, opened 2025-05-08
Network config fixes

gibmat:
* Default bond MAC address to first member's MAC address if not specified
* Fixes for VLAN configuration
* Don't show network interfaces without any configured address in TUI


stgraber:
@gibmat second time the tests fail on the same spot, so it looks like we may have broken networking in this PR somehow

gibmat:
> @gibmat second time the tests fail on the same spot, so it looks like we may have broken networking in this PR somehow

Yeah, the issue is with the default setting of
```
[BridgeVLAN]
VLAN=1-4094
```
without any vlan tagging happening.

stgraber (approved):

PR #119, opened 2025-05-22
Installer fixes

gibmat:
* Don't consider ramdisks or floppy devices
* Run `blkdiscard` to wipe target device
* Handle partially wiped disks that don't have a main GPT table
* Check if Secure Boot is enabled
* When system requirements check fails, don't enter a tight startup loop

### DIFF
--- a/incus-osd/internal/install/install.go
+++ b/incus-osd/internal/install/install.go
@@ -518,7 +518,7 @@ func copyPartitionDefinition(ctx context.Context, src string, tgt string, partit
 	partitionNameRegex := regexp.MustCompile(`Partition name: '(.+)'`)
 	partitionSizeRegex := regexp.MustCompile(`Partition size: \d+ sectors \((.+)\)`)
 
-	partitionHexCode := ""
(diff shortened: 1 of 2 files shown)


incus-osd/internal/install/install.go
```diff
@@ -315,12 +315,10 @@ func (i *Install) performInstall(ctx context.Context, modal *tui.Modal, sourceDe
...
-	if i.config.ForceInstall {
-		_, err := subprocess.RunCommandContext(ctx, "sgdisk", "-Z", targetDevice)
-		if err != nil {
-			return err
-		}
+	// Before starting the install, run blkdiscard to wipe target device.
+	_, err = subprocess.RunCommandContext(ctx, "blkdiscard", "-f", targetDevice)
+	if err != nil {
+		return err
 	}
```
stgraber: It's not guaranteed that all drive will handle `blkdiscard` so we shouldn't treat as fatal and always run the zap operation afterwards in case blkdiscard didn't work or wasn't sufficient.

PR #131, opened 2025-05-28
Fix missing seed detection when running from a CDROM

### LINKED ISSUE #125: Proxmox installation ??
Hi 
Trying to install on proxmox but always getting errors (Using ISO version). Is this even possible?
Proxmox VM settings: 
System : Graphics default
                Machine type q35 (Also tried with i440fx)
                BIOS UEFI
                EFI Storage selected (pre enroll keys disabled)
                TPM selected
Hard drive 260 GB (SCSI)
network Default 

Thank you very much for your great work on this and incus.

Note: I have incus running over ubuntu 22.04 on a VM on proxmox but i like to try it with incus os to check the speed difference.

gibmat:
Closes #125

### DIFF
--- a/incus-osd/internal/install/install.go
+++ b/incus-osd/internal/install/install.go
@@ -168,29 +168,13 @@ func (i *Install) DoInstall(ctx context.Context) error {
 	return i.rebootUponDeviceRemoval(ctx, sourceDevice)
 }
 
-// getSourceDevice determines the underlying device incus-osd is running on and if it is read-only.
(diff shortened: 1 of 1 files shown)


stgraber:
@gibmat the logic for "running from a cdrom" appears to be "is there a cdrom drive", which doesn't seem quite right :)

Can we tweak the check to actually ensure we're running from said cdrom?
I don't want a system booting from a USB stick but with a cdrom drive present to incorrectly trigger the logic.

stgraber (approved):

Issue #87, opened 2025-05-02
Implement network state


stgraber:
/1.0/system/network should report network state.

Probably something like:
```
state:
  interfaces:
    enp1:
      addresses:
        - addr1
        - addr2
      routes:
        - to: 1.2.3.4/24
          via: 1.1.1.1
      speed: 10000
      state: up
      lldp:
        - name: neighbor1
          chassis_id: XYZ
          port_id: XYZ
          port: 1/2/3 (may be missing)
        - name: neighbor2
          chassis_id: XYZ
          port_id: XYZ
          port: 1/2/3 (may be missing)
      stats:
        rx_bytes: 123
        tx_bytes: 123
        rx_errors: 0
        tx_errors: 0
    enp2:
      addresses:
        - addr1
        - addr2
      routes:
        - to: 1.2.3.4/24
          via: 1.1.1.1
      speed: 10000
      state: up
      lldp:
        - name: neighbor1
          chassis_id: XYZ
          port_id: XYZ
          port: 1/2/3 (may be missing)
        - name: neighbor2
          chassis_id: XYZ
          port_id: XYZ
          port: 1/2/3 (may be missing)
      stats:
        rx_bytes: 123
        tx_bytes: 123
        rx_errors: 0
        tx_errors: 0
    my-bond:
      addresses:
        - addr1
        - addr2
      routes:
        - to: 1.2.3.4/24
          via: 1.1.1.1
      speed: 10000
      state: up
      members:
        enp3:
          state: up
          local_mac: MAC
          remote_mac: MAC
          stats:
            rx_bytes: 123
            tx_bytes: 123
            rx_errors: 0
            tx_errors: 0
          lldp:
            - name: neighbor1
              chassis_id: XYZ
              port_id: XYZ
              port: 1/2/3 (may be missing)
            - name: neighbor2
              chassis_id: XYZ
              port_id: XYZ
              port: 1/2/3 (may be missing)
        enp4:
          state: up
          local_mac: MAC
          remote_mac: MAC
          stats:
            rx_bytes: 123
            tx_bytes: 123
            rx_errors: 0
            tx_errors: 0   
          lldp:
            - name: neighbor1 (may be missing)
              chassis_id: XYZ
              port_id: XYZ
              port: 1/2/3 (may be missing)              
            - neighbor2
              chassis_id: XYZ
              port_id: XYZ
              port: 1/2/3 (may be missing)
      stats:
        rx_bytes: 123
        tx_bytes: 123
[...]

PR #138, opened 2025-05-30
Implement network state REST endpoint

gibmat:
Return a representation of the system's configured network state when sending a `GET` request to the `/1.0/system/network` endpoint. I think this should pretty much cover what we need, although slight tweaking of the representation may be needed down the road.

Closes: #87


incus-osd/api/system_network.go
```diff
@@ -4,7 +4,7 @@ package api
 type SystemNetwork struct {
 	Config *SystemNetworkConfig `json:"config" yaml:"config"`
 
-	State struct{} `json:"state" yaml:"state"`
+	State SystemNetworkState `json:"-" yaml:"-"`
```
stgraber: Why is this in-lined? Our other endpoint have their state within a `state` key as was originally the case here.

incus-osd/api/system_network.go
```diff
@@ -85,3 +85,39 @@ type SystemNetworkProxy struct {
...
+	Addresses []string                               `json:"addresses,omitempty"  yaml:"addresses,omitempty"`
+	Routes    []SystemNetworkRoute                   `json:"routes,omitempty"     yaml:"routes,omitempty"`
+	MTU       int                                    `json:"mtu"                  yaml:"mtu"`
+	Speed     string                                 `json:"speed,omitempty"      yaml:"speed,omitempty"`
+	State     string                                 `json:"state"                yaml:"state"`
+	Stats     SystemNetworkInterfaceStats            `json:"stats"                yaml:"stats"`
+	LLDP      []SystemNetworkLLDPState               `json:"lldp,omitempty"       yaml:"lldp,omitempty"`
+	Members   map[string]SystemNetworkInterfaceState `json:"members,omitempty"    yaml:"members,omitempty"`
+	LocalMAC  string                                 `json:"local_mac,omitempty"  yaml:"local_mac,omitempty"`
+	RemoteMAC string                                 `json:"remote_mac,omitempty" yaml:"remote_mac,omitempty"`
```
stgraber: Might make sense to move that under an LACP struct and key.

incus-osd/internal/systemd/networkd.go
```diff
@@ -75,7 +81,13 @@ func ApplyNetworkConfiguration(ctx context.Context, networkCfg *api.SystemNetwor
...
 
 	// Wait for the network to apply.
-	return waitForNetworkOnline(ctx, networkCfg, timeout)
+	err = waitForNetworkOnline(ctx, networkCfg, timeout)
+	if err != nil {
+		return err
+	}
+
+	// Update the network state.
+	return UpdateNetworkState(ctx, n)
```
stgraber: What's the reason for updating it during apply?

I don't see us ever really wanting to access outdated information from the shared state, so only getting it on-demand would make sense.

incus-osd/api/system_network.go
gibmat: Network state is ephemeral, and we can always re-create it whenever needed, so I didn't think it made sense to write to disk as part of the daemon's state. If we really want to write it to disk I can easily add it back.

(`-` excludes the field from marshalling; `,inline` inlines the field.)

incus-osd/internal/systemd/networkd.go
gibmat: It was just to ensure we'd always have the network state reflect a new network configuration. If we want to rely on first calling `UpdateNetworkState()` whenever we want that information, I can remove that change.

incus-osd/api/system_network.go
stgraber: Right, but the problem is that `GET /1.0/system/network` should return `SystemNetwork`, so the struct with both the `config` and `state` members.

That's because we need to be able to do `GET /1.0/system/network`, modify the `config` part of it and then throw it back at the server through `PUT /1.0/system/network`.

With this PR, we end up having `GET` only return the state while `PUT` needs the full SystemNetwork struct.

gibmat:
@stgraber Not sure why the test keeps failing, although I noticed that the test run for the main branch after merging #139 also failed in the same manner

stgraber:
@gibmat there seems to be a problem with how we set up applications with the local provider.

The log shows "Downloading" as the last entry after 1 min wait since restarting `incus-osd`.
Since we're on the local provider, download should be instant, so no idea why we didn't hit installation and provisioning...

PR #140, opened 2025-06-02
Trigger updates via API

gibmat:
Trigger OS and application updates via the API, rather than restarting the incus-osd daemon. This will more closely reflect how updates will be applied on production systems.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -184,4 +184,9 @@ jobs:
           sleep 3m
 
           incus exec test-incus-os -- journalctl -u incus-osd -b0
-          incus exec test-incus-os -- grep IMAGE_VERSION /usr/lib/os-release
+
+          incus restart test-incus-os
+
+          sleep 1m
+
+          incus exec test-incus-os -- grep $RELEASE /usr/lib/os-release


mkosi.images/base/mkosi.conf.d/03-core-packages.conf
```diff
@@ -2,6 +2,7 @@
 Packages=
     apparmor
     dbus
+    curl
```
stgraber, line 4: Not alphabetically sorted :)

stgraber: Don't we have `curl` in the `debug` application too? If we do, we should drop it from there since it's now in base.

stgraber: I'd probably move that change to its own commit too

incus-osd/cmd/incus-osd/main.go
```diff
@@ -508,11 +519,16 @@ func updateChecker(ctx context.Context, s *state.State, t *tui.TUI, p providers.
...
-			// Reload the application.
+			// Start/reload the application.
 			if !isStartupCheck {
-				slog.Info("Reloading application", "name", appName, "version", appVersion)
+				if app.IsRunning(ctx) {
+					slog.Info("Reloading application", "name", appName, "version", appVersion)
+
+					err = app.Update(ctx, appVersion)
+				} else {
+					err = startInitializeApplication(ctx, s, appName)
```
stgraber: nit but I prefer to handle the error immediately after it's received, the current syntax here makes it easy to forget that the error is handled in the parent block and cause it to get overridden

PR #144, opened 2025-06-04
Implement a flasher tool

### LINKED ISSUE #70: Implement flasher tool
We should provide a command line tool which can:
 - Figure out the latest version of Incus OS
 - Fetch it from Github
 - Provide a basic customization menu, offering:
   - 1) Toggle between install or running mode (when switching to install mode, prompt if a specific drive should be used)
   - 2) Select applications
   - 3) Provide network seed data
   - 4) Provide Incus seed data (if Incus in application list)
   - 0) Write the image (asks about the path to write to and then writes the image with correct seed)

We can then have `scripts/spawn-image` call that script to handle the downloading and customization piece but passing a pre-defined target image path through an env variable to not have it prompt about the target location.

### LINKED ISSUE #130: Document basic installation and usage
We should add some documentation covering basic installation and usage of Incus OS so people don't have to rely on the `spawn-image` script:

- System requirements
- How to configure and inject seed configuration
- How to connect to Incus after Incus OS install is finished

gibmat:
Closes #70, Closes #130

### DIFF
--- a/doc/install-seed.md
+++ b/doc/install-seed.md
@@ -1,0 +1,65 @@
+# Install Seed
+Incus OS depends on an "install seed" to automate the installation process. Most
+users should use the [flasher tool](flasher-tool.md) which provides a simple way
+to configure the install seed without requiring detailed understanding of the
+technical details below.
+
+## Format and location
+The install seed is a simple tar archive consisting of one or more json/yaml
+configuration files. The tar file is written directly to the start of the second
+partition of the install image. At runtime, Incus OS will attempt to read the
+install seed from the second partition and use any data present during the
+install process.
+
+## Seed contents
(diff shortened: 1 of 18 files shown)


doc/install-seed.md
```diff
@@ -0,0 +1,59 @@
...
+technical details below.
+
+## Format and location
+The install seed is a simple tar archive consisting of one or more json/yaml
+configuration files. The tar file is written directly to the start of the second
+partition of the install image. At runtime, Incus OS will attempt to read the
+install seed from the second partition and use any data present during the
+install process.
+
+## Seed contents
```
stgraber, line 14: We're missing the provider seed in that list

stgraber (approved):

Issue #133, opened 2025-05-28
Investigate secureboot key rotation


stgraber:
We need to come up with a solid plan to handle trusting a new secureboot key and eventually revoking the old one.

Ideally we'd have a root key that's kept offline which signs an intermediate key that's typically used for a year before we switch over to the next key and let the previous key expire.

That works pretty well with TLS but secureboot is a bit of a different beast so we need to do some investigation there.

PR #150, opened 2025-06-09
Secure Boot key rotation

gibmat:
This PR implements the functionality needed to update Secure Boot keys (KEK, db, and dbx) as well as handle rotation of the Secure Boot key used to sign the UKIs.

Closes: #133


mkosi.images/base/mkosi.conf.d/90-certificates.sh
```diff
@@ -4,4 +4,7 @@
 mkdir -p "${DESTDIR}/usr/share/certs"
 cp /etc/ssl/certs/ca-certificates.crt "${DESTDIR}/usr/share/certs/"
 
+# Append Incus OS CAs.
+cat /work/src/certs/cas/*.crt >> "${DESTDIR}/usr/share/certs/ca-certificates.crt"
```
stgraber: What's the reason for this one?

On the image server side, we'll have regular TLS for https://images.linuxcontainers.org so Go-TLS will be happy with the normal ca-certificates.

Then the update metadata itself will have a detached signature (`update.json.sig`) which will come from the current year's update CA and can be validated through direct invocations of Go TLS. So we'll effectively have a `.crt` chain and `.crl` as part of the image which we'll use to validate the update signatures all the way back to our root certificate. This validation will be done entirely separate from ca-certificates.crt so we don't end up trusting something signed by some other random certificate.

incus-osd/internal/systemd/cryptenroll.go
```diff
@@ -13,30 +13,41 @@ import (
...
+	// First, generate a recovery key for the root volume.
+	recoveryPassword, err := subprocess.RunCommandContext(ctx, "systemd-cryptenroll", "--unlock-tpm2-device", "auto", "--recovery-key", "/dev/disk/by-partlabel/root-x86-64")
 	if err != nil {
 		return err
 	}
+	recoveryPassword = strings.TrimSuffix(recoveryPassword, "\n")
 
-	s.System.Encryption.Config.RecoveryKeys = append(s.System.Encryption.Config.RecoveryKeys, strings.TrimSuffix(output, "\n"))
+	// Second, set the same recovery key for the swap volume. Need to pass to systemd-cryptenroll via NEWPASSWORD environment variable.
+	_, _, err = subprocess.RunCommandSplit(ctx, append(os.Environ(), "NEWPASSWORD="+recoveryPassword), nil, "systemd-cryptenroll", "--unlock-tpm2-device", "auto", "--password", "/dev/disk/by-partlabel/swap")
```
gibmat, line 26: This will eventually break any systems already installed, since their swap volume won't be able to be updated when a new Secure Boot key is deployed.

incus-osd/internal/systemd/sysupdate.go
```diff
@@ -53,13 +56,41 @@ func GetCurrentRelease(_ context.Context) (string, string, error) { //nolint:rev
...
+		}
+	}
+
+	secureBootKeyChanged, err := secureboot.UKIHasDifferentSecureBootCertificate(ctx, newUKIFile)
+	if err != nil {
+		return err
+	}
+
+	if secureBootKeyChanged {
+		err := secureboot.HandleSecureBootKeyChange(ctx, luksPassword, newUKIFile, newUsrImageFile)
```
gibmat, line 88: This call takes a while; it would be nice to provide some sort of updates to the user.

incus-osd/internal/providers/provider_github.go
```diff
@@ -53,6 +53,11 @@ func (*github) Type() string {
 	return "github"
 }
 
+func (*github) GetSecureBootCertUpdate(_ context.Context) (SecureBootCertUpdate, error) {
+	// TODO -- query GitHub for current list of Secure Boot key updates.
+	return nil, ErrNoUpdateAvailable
```
gibmat, line 57: Need to figure out how we want to handle distributing secure boot key updates via GitHub. Since updates will be signed offline independent of the normal release workflow, my first thought was just add each update to a special directory (`.github/secure-boot-updates/`?) as a normal commit. We ought to be able to get a simple directory listing then, and the code could parse that for potential updates.

incus-osd/internal/providers/provider_operations_center.go
```diff
@@ -147,6 +147,11 @@ func (*operationsCenter) Type() string {
 	return "operations-center"
 }
 
+func (*operationsCenter) GetSecureBootCertUpdate(_ context.Context) (SecureBootCertUpdate, error) {
+	// TODO -- query OperationsCenter for current list of Secure Boot key updates.
+	return nil, ErrNoUpdateAvailable
```
gibmat, line 152: I expect there will be some API provided by OperationsCenter that will return any available secure boot key updates.

Makefile
```diff
@@ -34,17 +34,31 @@ initrd-deb-package:
 static-analysis:
 	(cd incus-osd && golangci-lint run)
 
-.PHONY: build
-build: incus-osd flasher-tool initrd-deb-package
-	-mkosi genkey
+.PHONY: generate-test-certs
+generate-test-certs:
```
gibmat, line 43: This will need some tweaking to properly integrate with the GitHub release process.

Makefile
```diff
@@ -34,17 +34,31 @@ initrd-deb-package:
...
+.PHONY: build
+build: generate-test-certs incus-osd flasher-tool initrd-deb-package
 	mkdir -p mkosi.images/base/mkosi.extra/usr/local/bin/
 	cp incus-osd/incus-osd mkosi.images/base/mkosi.extra/usr/local/bin/
 	sudo rm -Rf mkosi.output/base* mkosi.output/debug* mkosi.output/incus*
 	sudo -E $(shell command -v mkosi) --cache-dir .cache/ build
 	sudo chown $(shell id -u):$(shell id -g) mkosi.output
 
+	# For some reason getting the image name via $(shell ...) is always empty here?
+	sudo ./scripts/inject-secure-boot-vars.sh `ls mkosi.output/IncusOS_*.raw | grep -v usr | grep -v esp | sort | tail -1`
```
gibmat, line 68: After generating the new signing keys, we'll have to import the resulting `*.auth` files somewhere into the repo, then adjust the script to pick them up. The actual secure boot signing key in use will continue to be provided via a workflow secret.

scripts/test/generate-test-certificates.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
stgraber: I'm not seeing any bashism in there, should be `#!/bin/sh` then

Makefile
stgraber: We probably shouldn't have `build` depend on `generate-test-certs` and just make that an optional step. That way production can just hit `make build` and use the normal mkosi.crt and mkosi.key on disk.

We probably also should make `build` run `mkosi genkey` (old logic) if `mkosi.crt` or `mkosi.key` doesn't exist.

That way `make` out of the box works the same as it does today, `make generate-test-certs && make` works for testing the full PKI and production can keep using its hardcoded `mkosi.crt` and `mkosi.key` with `make`.

Makefile
stgraber: We'll want this to be conditional on `certs/` containing something, right?

incus-osd/internal/secureboot/secureboot.go
```diff
@@ -0,0 +1,357 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"os"
+	"slices"
+
+	"github.com/google/go-eventlog/register"
+	"github.com/google/go-eventlog/tcg"
+	"github.com/smallstep/pkcs7"
```
stgraber: go-eventlog is Apache2
pkcs7 is MIT

We're good on the licensing front

incus-osd/internal/systemd/cryptenroll.go
```diff
@@ -13,30 +13,41 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	// Finally, save the recovery key into the state.
+	s.System.Encryption.Config.RecoveryKeys = append(s.System.Encryption.Config.RecoveryKeys, recoveryPassword)
```
stgraber, line 32: What happens when the user modifies the recovery keys?
Do we allow them to remove all keys? (sounds like we really shouldn't)

incus-osd/internal/systemd/cryptenroll.go
```diff
@@ -45,18 +56,24 @@ func AddEncryptionKey(ctx context.Context, s *state.State, key string) error {
...
 // Due to systemd-cryptenroll only being able to wipe slots by index or type, we must first
 // remove all recovery and password slots, then re-add any remaining keys.
 func DeleteEncryptionKey(ctx context.Context, s *state.State, key string) error {
 	if !slices.Contains(s.System.Encryption.Config.RecoveryKeys, key) {
 		return errors.New("provided encryption key is not enrolled")
 	}
 
+	if len(s.System.Encryption.Config.RecoveryKeys) == 1 {
+		return errors.New("cannot remove only existing recovery key")
+	}
```
stgraber, line 69: Well, that answers that question ;)

incus-osd/internal/secureboot/secureboot.go
```diff
@@ -16,11 +17,73 @@ import (
...
+// to the current EFI value, and then updates the expected PCR7 value used to
+// decrypt the root file system and swap at boot.
+func AppendEFIVarUpdate(ctx context.Context, efiUpdateFile string, varName string) error {
+	// Verify the file exists.
+	_, err := os.Stat(efiUpdateFile)
+	if err != nil {
+		return err
+	}
+
+	// TODO -- when applying dbx, verify neither of the two images are relying on that certificate to boot.
```
stgraber: Do we need to address that in this PR?

incus-osd/internal/secureboot/secureboot.go
```diff
@@ -16,11 +17,73 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	err = validateUntrustedTPMEventLog(eventLog)
+	if err != nil {
+		return err
+	}
+
+	// By default, sysfs mounts EFI variables with the immutable attribute set. We need to remove it prior to appending the update.
```
stgraber, line 154: Do we need to restore the immutable flag afterwards? (defer)

mkosi.images/base/mkosi.conf.d/03-core-packages.conf
```diff
@@ -5,6 +5,7 @@ Packages=
     dbus
     dosfstools
     e2fsprogs
+    efitools
```
stgraber, line 8: Should be moved to own commit (`base: Add efitools`)

incus-osd/internal/providers/provider_github.go
```diff
@@ -53,6 +53,11 @@ func (*github) Type() string {
 	return "github"
 }
 
+func (*github) GetSecureBootCertUpdate(_ context.Context) (SecureBootCertUpdate, error) {
+	// TODO -- query GitHub for current list of Secure Boot key updates.
```
stgraber: I'd remove the TODO, this provider is going to die before we cut over to the new signing keys anyway :)

incus-osd/internal/providers/provider_local.go
```diff
@@ -29,7 +29,7 @@ func (*local) ClearCache(_ context.Context) error {
 }
 
 func (*local) RefreshRegister(_ context.Context) error {
-	// No registration with the Github provider.
+	// No registration with the local provider.
```
stgraber, line 32: Can you push a separate commit fixing the comments since that's not the only occurrence of this one.

incus-osd/internal/providers/provider_operations_center.go
stgraber: Yeah, I expect we'll first switch to using our image servers to distribute our updates and as part of the filelist within an update, we'll have UEFI PKI update files.

So in general, we'll be able to pick the latest set of SecureBoot PKI variables and keep track of what's applied somewhat similarly to applications.

incus-osd/internal/providers/structs.go
```diff
@@ -22,12 +22,19 @@ type OSUpdate interface {
 	Download(ctx context.Context, osName string, targetPath string, progressFunc func(float64)) error
 }
 
+// SecureBootCertUpdate represents a Secure Boot UEFI certificate update (typically a db or dbx addition).
+type SecureBootCertUpdate interface {
+	AvailableUpdates() map[string][]string
```
stgraber: Not sure what's that one about.

I would have expected maybe:
 - Version() string (version of the secureboot update, to be tracked in state so we don't re-apply needlessly)
 - IsNewerThan(otherVersion string) bool (for update checking)
 - Download(ctx contesxt.Context, targetPath string, progressFunc func(float64)) error (to dump all the files in the target folder)

We'd then have each SecureBoot update include always the same set of files, no partial updates. We want to be able to grab the latest secureboot update and apply it, not have to go through the last X entries and see which one has just a db update, which one has latest dbx, ...

incus-osd/cmd/incus-osd/main.go
```diff
@@ -351,7 +357,7 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error {
 		}
 	}
 
-	// Set up handler for shutdown tasks.
+	// Set up handler for daemon actions.
```
stgraber, line 362: I'd split that as its own commit

incus-osd/cmd/incus-osd/main.go
```diff
@@ -502,6 +520,8 @@ func updateChecker(ctx context.Context, s *state.State, t *tui.TUI, p providers.
 				modal = t.AddModal(s.OS.Name + " Update")
 			}
 			modal.Update(s.OS.Name + " has been updated to version " + newInstalledOSVersion + ".\nPlease reboot the system to finalize update.")
+
+			s.RebootRequired = true
```
stgraber, line 526: I'd make a standalone commit adding the concept of RebootRequired

Makefile
```diff
@@ -155,6 +155,16 @@ test-update:
 
 	incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system -X PUT -d '{"action": "update"}'
 
+.PHONY: test-update-sb-keys
```
stgraber, line 167: I'd move the testing as a follow-up commit to this change

incus-osd/internal/secureboot/secureboot.go
stgraber: Yes, addressed later.

incus-osd/internal/providers/provider_github.go
stgraber: For Github, we simply won't. We'll drop that provider soon.

stgraber:
As discussed, we should switch the secureboot update to be a single tarball (kinda similar to an application image) and have it include trees for both db and dbx with individual files for each change in there.

incus-osd/internal/secureboot/secureboot.go
gibmat: No; when the system reboots `sysfs` will automatically apply the immutable attribute again.

gibmat:
@stgraber this should be pretty much ready.

scripts/inject-secure-boot-vars.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
stgraber: Not seeing any bashism, can that be /bin/sh?

incus-osd/cmd/incus-osd/main.go
```diff
@@ -227,8 +227,9 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error {
 		slog.Debug("Platform keyring entry", "name", key.Description, "key", key.Fingerprint)
 	}
 
-	// If no encryption recovery keys have been defined for the root partition, generate one before going any further.
+	// If no encryption recovery keys have been defined for the root and swap partitions, generate one before going any further.
 	if len(s.System.Encryption.Config.RecoveryKeys) == 0 {
+		slog.Info("First boot: auto-generating encryption recovery key, this may take a few seconds")
```
stgraber: I don't think we need the `First boot: ` prefix as it will look weird if we have multiple ones of those. We can probably just log `Auto-generating encryption recovery key` so we know that something is going on.

stgraber: (trying to keep it short so we don't have it wrap :))

incus-osd/internal/secureboot/secureboot.go
```diff
@@ -13,16 +13,93 @@ import (
...
+		return err
+	}
+
+	err = validateUntrustedTPMEventLog(eventLog)
+	if err != nil {
+		return err
+	}
+
+	// Part 1 -- Update the systemd-boot EFI stub.
+	{
```
stgraber: What's the purpose of that block? Does that scope the defers?

incus-osd/internal/secureboot/secureboot.go
```diff
@@ -13,16 +13,93 @@ import (
...
+
+		_, err = subprocess.RunCommandContext(ctx, "cp", filepath.Join(mountDir, "lib/systemd/boot/efi/systemd-bootx64.efi.signed"), "/boot/EFI/systemd/systemd-bootx64.efi")
+		if err != nil {
+			return err
+		}
+
+		_, err = subprocess.RunCommandContext(ctx, "cp", filepath.Join(mountDir, "lib/systemd/boot/efi/systemd-bootx64.efi.signed"), "/boot/EFI/BOOT/BOOTX64.EFI")
+		if err != nil {
+			return err
+		}
```
stgraber: Just worth keeping in mind that we'll need to tweak that a bit when we add arm64 support as BOOTX64.EFI turns into BOOTAA64.EFI or something like that

incus-osd/internal/providers/provider_github.go
```diff
@@ -237,7 +241,7 @@ func (p *github) downloadAsset(ctx context.Context, assetID int64, target string
 		}
 
 		// Update progress every 24MiB.
-		if count%6 == 0 {
+		if progressFunc != nil && count%6 == 0 {
```
stgraber, line 244: Unrelated to this commit, should be split out

incus-osd/internal/providers/provider_local.go
```diff
@@ -189,7 +221,7 @@ func (p *local) copyAsset(_ context.Context, name string, target string, progres
 		}
 
 		// Update progress every 24MiB.
-		if count%6 == 0 {
+		if progressFunc != nil && count%6 == 0 {
```
stgraber, line 224: Same as above, should be in a separate commit

incus-osd/internal/providers/provider_operations_center.go
```diff
@@ -453,7 +458,7 @@ func (p *operationsCenter) downloadAsset(ctx context.Context, assetURL string, t
 		}
 
 		// Update progress every 24MiB.
-		if count%6 == 0 {
+		if progressFunc != nil && count%6 == 0 {
```
stgraber, line 461: This one too

incus-osd/cmd/incus-osd/main.go
```diff
@@ -693,3 +713,159 @@ func checkDoAppUpdate(ctx context.Context, s *state.State, t *tui.TUI, p provide
 
 	return "", nil
 }
+
+func checkDoSecureBootCertUpdate(ctx context.Context, s *state.State, t *tui.TUI, p providers.Provider, isStartupCheck bool) error {
+	slog.Debug("Checking for Secure Boot key updates")
+
+	if s.RebootRequired {
+		slog.Warn("A reboot of the system is required to finalize a pending update")
```
stgraber: Won't that get logged in a loop (every update check)?

incus-osd/cmd/incus-osd/main.go
```diff
@@ -693,3 +713,159 @@ func checkDoAppUpdate(ctx context.Context, s *state.State, t *tui.TUI, p provide
...
+		}
+
+		err = applyIndividualSecureBootUpdates(ctx, s, t, tmpDir, isStartupCheck)
+		if err != nil {
+			return err
+		}
+
+		// If an EFI variable was updated, we'll either be rebooting automatically or waiting
+		// for the user to restart the system before going any further.
+		if s.RebootRequired {
```
stgraber, line 782: Should we log something in this case?

stgraber:
Another round of very minor nits and we'll be able to merge this one.

scripts/inject-secure-boot-vars.sh
gibmat: `checkbashisms` flagged the use of `$EUID`, which I replaced.

incus-osd/internal/secureboot/secureboot.go
gibmat: That had been my intention, but defer scoping is only done at the function level. Moved that code to its own helper function.

incus-osd/internal/secureboot/secureboot.go
gibmat: Added a helper function to return the correct paths depending on architecture, currently only supporting amd64.

incus-osd/cmd/incus-osd/main.go
gibmat: It will yes; I'll bump it to just a debug level log message so we can at least are logging that the system is in need of a reboot.

stgraber (approved):

PR #177, opened 2025-06-26
Add URL for GitHub provider to fetch Secure Boot keys update

gibmat:
Depends on #175 being part of a tagged release first.


stgraber:
@gibmat http://images.linuxcontainers.org/os/IncusOS_SB_transition.tar.gz is now live

gibmat:
Tested with the current release (202506220351) from a fresh install and the updated Secure Boot keys applied correctly. Will do a little more testing from an older release to verify the upgrade path works as expected.

gibmat:
Successfully started from an install of 202506170021, upgraded to 202506220351, then ran a locally-built `incus-osd` which added missing recovery key to swap volume and appended the two Secure Boot key updates.

PR #187, opened 2025-07-02
Refactor and extend security API

### LINKED ISSUE #179: Add an "encryption recovery" API
If something causes auto-unlocking of the LUKS volumes to fail, the recovery passphrase can be used to boot the system, but there's currently no easy way to then cleanup the underlying issue and allow the system to auto-unlock once again.

We should add an "encryption recovery" API that looks at each encrypted volume:
* Flat out refuse to do anything if:
  - Secure Boot is disabled (so we can't be tricked into unlocking under an attacker's environment)
  - The current TPM state can be used to unlock a given volume (no reason to do anything)
* Rebind the current PCR7 hash for the volume's encryption
* Update PCR11 public key binding, if needed
* Immediately reboot the system

As a special case, if the swap volume won't auto-unlock, we can simply blow it away and re-create an encrypted volume with the current good state.

### LINKED ISSUE #180: Track state of encrypted volumes
We should record the state of each encrypted volume (did it auto-unlock, was a recovery password used, unlock failed, etc). We should also track if we're happy with the Secure Boot and TPM states.

### LINKED ISSUE #181: Refactor encryption API
Currently the API endpoint is `/1.0/system/encryption`; it would probably make more sense to be `/1.0/system/security`.

### LINKED ISSUE #183: Add API to query Secure Boot certificates
Add an API that will return information about currently installed Secure Boot keys: type, fingerprint, CN, Issuer, etc.

gibmat:
Refactor and extend the security API.

Also includes logic to handle changes to the state struct, both via the new upgrade function approach and a temporary bit of logic to pre-mangle existing json input prior to its conversion to the new state format.

Closes #179, Closes #180, Closes #181, Closes #183


gibmat:
Getting annoyed with the linter and its non-deterministic behavior... exact same version locally inverts the two current "failures". :angry:

stgraber (approved):

PR #193, opened 2025-07-09
incus-osd/systemd/networkd: Fill in Hwaddr in state

stgraber:


stgraber:
@breml

incus-osd/.golangci.yml
```diff
@@ -56,6 +56,8 @@ linters:
           disabled: true
         - name: nested-structs
           disabled: true
+        - name: var-naming
+          disabled: true
```
breml, line 60: Me personally, I would not disable this in general but on a per use case.

stgraber: That was our first approach but it didn't work. 

When the issue is about the whole package name (in this case, util), golangci will report the package line of one of the files as being the problem. 

Exactly what file is random, so the more file you have in the package the less likely you are to get lucky and have CI pass...

Basically if we're lucky, golangci-lint would get us this warning in the same file as we have the nolint comment, then we're good. 

But otherwise it would issue two warnings, one about an unused nolint statement and one for the package name, through a file that doesn't have the nolint...

Technically that feels like a bug in golangci-lint/revive as you'd expect the behavior to be reproducible, but it currently isn't...

PR #210, opened 2025-07-29
Add storage API

### LINKED ISSUE #145: Define a storage API
We need an API for local storage:
* Report each local disk (zpool or unassigned, health, etc)
* Create/update/delete zpools
* Add/remove drives from zpools
* Status of local storage, including degraded zpools

gibmat:
Adds a preliminary storage API:

- `/1.0/system/storage`:
  - GET returns details about each drive and the status of each zpool
  - PUT creates/updates a new zpool
  - DELETE destroys a zpool and runs `blkdiscard` on its former members
- `/1.0/system/security`:
  - GET also returns base64 encoded zpool encryption keys

Closes #145


incus-osd/internal/util/disk.go
```diff
@@ -0,0 +1,77 @@
+package util
```
stgraber: Why util instead of the storage package?

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,49 @@
+package api
+
+import (
+	"encoding/json"
+)
+
+// SystemStorage defines a struct to hold information about the system's local storage.
+type SystemStorage struct {
+	State struct {
+		SMARTStatus map[string]json.RawMessage `json:"smart_status" yaml:"smart_status"`
```
stgraber: It'll be more work, but I think we should try to abstract things a bit more.

Basically having a `Drives []SystemStorageDrive` with each drive then including at least:
 - ID
 - Manufacturer
 - Model
 - Serial number
 - Bus
 - Removable
 - WWN
 - SMART status
 - Pool

We'll also want to do a bit of filtering.
For example, if we find a multipath device, we should only show the ID for the multipath device itself and exclude all of the individual drives that are part of it.

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,49 @@
+package api
+
+import (
+	"encoding/json"
+)
+
+// SystemStorage defines a struct to hold information about the system's local storage.
+type SystemStorage struct {
```
stgraber, line 4: I think defining and updating pools should be done through the `Config` struct same we do elsewhere.

So having something like:
```
Config {
    Pools [
        Pool {
            Name: "local-nvme",
            Type: "zfs-raidz1",
            Devices: ["dev1", "dev2"],
        }
        Pool {
            Name: "local-nvme",
            Type: "zfs-raidz2",
            Devices: ["dev1", "dev2", "dev3", "dev4", "dev5"],
            Cache: ["dev6", "dev7"],
            Log: ["dev8", "dev9"],
        }
    ]
}
```

We'd define those types for now:
 - zfs-raid0 (stripe across all devices)
 - zfs-raid1 (mirror across all devices)
 - zfs-raid10 (stripe across pairs of devices)
 - zfs-raidz1
 - zfs-raidz2
 - zfs-raidz3

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	ZpoolName string `json:"zpool_name" yaml:"zpool_name"`
+
+	// Type of the zpool: single, mirror, raidz1, raidz2, raidz3
+	Type string `json:"type" yaml:"type"`
+
+	// List of devices to be used when creating the pool.
+	Devices []string `json:"devices" yaml:"devices"`
+
+	// Optionally, a user-provided 32 byte raw encryption key, base64 encoded. If not provided, a random key will be generated.
+	RawEncryptionKey string `json:"raw_encryption_key,omitempty" yaml:"raw_encryption_key,omitempty"`
```
stgraber: I'd omit that for now, I think we should just always generate our own per-pool key and store it internally. Having the user provide their key is just likely to make things less safe.

If we ever get into the business of importing/exporting pools between systems, then we'll something like that, but it's not something we should look into now.

incus-osd/api/system_storage.go
stgraber: Replacement operations would then literally be a PUT to the config, replacing one drive ID with another. Same would go for device addition or removal (when supported).

incus-osd/internal/util/disk.go
gibmat: I suppose that makes more sense.

incus-osd/api/system_storage.go
gibmat: I think I've got a pretty good solution for this, although it will be interesting to see if any fields are missing/incorrect across wider hardware.

`lsblk` has a merge option (`-M`), which supposedly should handle cases like multipath devices and only show the actual device, but we'll need to test that.

stgraber:
Looks like there's a small test migration related to the state

stgraber:
Static analysis isn't happy ;)

gibmat:
@stgraber I think this should be ready now.

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,49 @@
...
+
+// SystemStorage defines a struct to hold information about the system's local storage.
+type SystemStorage struct {
+	Config struct {
+		Pools []SystemStoragePool `json:"pools" yaml:"pools"`
+	} `json:"config" yaml:"config"`
+
+	State struct {
+		Drives []SystemStorageDrive       `json:"drives" yaml:"drives"`
+		Zpools map[string]json.RawMessage `json:"zpools" yaml:"zpools"`
```
stgraber: I think we should have that be `Pools` and use a struct that we own (similar to what we did for network state). We're going to have 3rd party software hitting those APIs, so we don't really want to just pass through JSON structs from 3rd party tools which may be changing on us.

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,49 @@
...
+	ID              string                 `json:"id"                yaml:"id"`
+	ModelFamily     string                 `json:"model_family"      yaml:"model_family"`
+	ModelName       string                 `json:"model_name"        yaml:"model_name"`
+	SerialNumber    string                 `json:"serial_number"     yaml:"serial_number"`
+	Bus             string                 `json:"bus"               yaml:"bus"`
+	CapacityInBytes int                    `json:"capacity_in_bytes" yaml:"capacity_in_bytes"`
+	Removable       bool                   `json:"removable"         yaml:"removable"`
+	WWN             *SystemStorageDriveWWN `json:"wwn,omitempty"     yaml:"wwn,omitempty"`
+	SMARTSupport    json.RawMessage        `json:"smart_support"     yaml:"smart_support"`
+	SMARTStatus     json.RawMessage        `json:"smart_status"      yaml:"smart_status"`
```
stgraber: Same note about passing through raw JSON structs. I think we should just have a `SMART` key here which is a struct that we own. If it's `nil` then we don't have SMART data, if it's set, then we expose what we see as the most useful data rather than just dumping the whole thing at the user.

incus-osd/api/system_security.go
```diff
@@ -12,6 +12,7 @@ type SystemSecurity struct {
 		SecureBootEnabled               bool                                  `json:"secure_boot_enabled"                yaml:"secure_boot_enabled"`
 		SecureBootCertificates          []SystemSecuritySecureBootCertificate `json:"secure_boot_certificates"           yaml:"secure_boot_certificates"`
 		TPMStatus                       string                                `json:"tpm_status"                         yaml:"tpm_status"`
+		ZpoolEncryptionKeys             map[string]string                     `json:"zpool_encryption_keys"              yaml:"zpool_encryption_keys"`
```
stgraber: I think I'd go with `PoolRecoveryKeys` for that one.
We may at some point support encryption of LVM pools, whether local or shared which would then also have their recovery keys.

incus-osd/api/system_storage.go
gibmat: I've extended the `SystemStoragePool` struct to have a couple more fields that can be used for reporting a pool's status and lists of any degraded devices. It didn't make much sense to introduce another new struct that would be almost identical to the existing one.

incus-osd/internal/tui/tui.go
```diff
@@ -267,7 +267,7 @@ func (t *TUI) redrawScreen() {
 		}
 
 		if !t.state.System.Security.State.EncryptionRecoveryKeysRetrieved {
-			t.frame.AddText("WARNING: Encryption recovery key has not been retrieved yet!", false, tview.AlignLeft, tcell.ColorRed)
+			t.frame.AddText("WARNING: Encryption recovery and/or zpool encryption keys have not yet been retrieved!", false, tview.AlignLeft, tcell.ColorRed)
```
stgraber: `WARNING: Some encryption recovery keys have not been retrieved yet!` to keep this a bit more generic?

incus-osd/api/system_storage.go
```diff
@@ -0,0 +1,59 @@
...
+// SystemStorageDrive defines a struct that holds information about a specific drive.
+type SystemStorageDrive struct {
+	ID              string                   `json:"id"                    yaml:"id"`
+	ModelFamily     string                   `json:"model_family"          yaml:"model_family"`
+	ModelName       string                   `json:"model_name"            yaml:"model_name"`
+	SerialNumber    string                   `json:"serial_number"         yaml:"serial_number"`
+	Bus             string                   `json:"bus"                   yaml:"bus"`
+	CapacityInBytes int                      `json:"capacity_in_bytes"     yaml:"capacity_in_bytes"`
+	Removable       bool                     `json:"removable"             yaml:"removable"`
+	WWN             *SystemStorageDriveWWN   `json:"wwn,omitempty"         yaml:"wwn,omitempty"`
```
stgraber: Can we just have it as a string using the usual `0xXYZ` notation? Example would be `0x50014ee20ce3848a`
That's what we take as input in the Multipath config.

incus-osd/internal/tui/tui.go
gibmat: Done

incus-osd/api/system_storage.go
gibmat: It could be a string if you can point me to the correct way to construct it based off of the naa, oui, and id fields, which are what `smartctl` returns.

gibmat: Think I've got it figured out.

stgraber (approved):

PR #220, opened 2025-08-05
Fix bringing default network online

### LINKED ISSUE #207: Don't require all network interfaces to be up if no network seed is provided
When booting without any network seed configuration, we currently require all detected network interfaces to be up before continuing. We should require only one.

(From https://discuss.linuxcontainers.org/t/debug-mode-in-incus-os-baremetal/24160/33)

gibmat:
When no network seed is provided, don't require any auto-generated interfaces to be online, but rather rely on DNS resolution to indicate when the network is up.

Closes #207

### DIFF
--- a/incus-osd/internal/seed/network.go
+++ b/incus-osd/internal/seed/network.go
@@ -103,10 +103,14 @@ func getDefaultNetworkConfig() (*api.SystemNetworkConfig, error) {
 			continue
 		}
 
+		// Auto-generated interfaces set RequiredForOnline to "no", and we rely
+		// on the DNS check after interface configuration to indicate when there's
+		// good network connectivity.
 		ret.Interfaces = append(ret.Interfaces, api.SystemNetworkInterface{
-			Name:      i.Name,
-			Hwaddr:    i.HardwareAddr.String(),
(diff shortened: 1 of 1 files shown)


incus-osd/api/system_network.go
```diff
@@ -48,6 +48,7 @@ type SystemNetworkInterface struct {
 	Hwaddr            string               `json:"hwaddr"                        yaml:"hwaddr"`
 	Roles             []string             `json:"roles,omitempty"               yaml:"roles,omitempty"`
 	LLDP              bool                 `json:"lldp"                          yaml:"lldp"`
+	AutoGenerated     bool                 `json:"auto_generated"                yaml:"auto_generated"`
```
stgraber: I don't really love this approach. It's a property that's going to be user-facing, which is meant to keep track of something internal but can still be set by the user.

Wouldn't it be sufficient to generate all interfaces with `RequiredForOnline=false`?
We have the DNS test right afterwards which would basically require one of the NICs to be functional anyway.

gibmat: I had been relying on just the state of the interfaces, but checking on working DNS resolution can also work.

PR #222, opened 2025-08-06
Fix network device naming

### LINKED ISSUE #218: Seed NIC name not picked up in incus
# Context

This is my network seed:

```
interfaces:
  - name: mgt
    addresses:
      - dhcp4
    hwaddr: C4:CB:E1:XX:YY:ZZ

    roles:
      - management
      - instances
```

# What works

When booted, the TUI Incus OS displays reports:

```
mgt(<ip address>)
```

as expected.

# What doesn't

```
$ incus network list

+----------------+----------+---------+---------------+---------------------------+----------------------------+---------+---------+
|      NAME      |   TYPE   | MANAGED |     IPV4      |           IPV6            |        DESCRIPTION         | USED BY |  STATE  |
+----------------+----------+---------+---------------+---------------------------+----------------------------+---------+---------+
| enc4cbe1axxyyzz | physical | NO      |               |                           |                            | 0       |         |
+----------------+----------+---------+---------------+---------------------------+----------------------------+---------+---------+
...
```

There is probably a udev rule here that overwrote this, changing `mgt` into `enc4cbe1axxyyzz`. The rest is named using the usual permanent naming scheme like `eno8403` and so on.

gibmat:
Closes #218


incus-osd/internal/systemd/networkd_validate.go
```diff
@@ -200,6 +201,14 @@ func validateName(name string) error {
 		return errors.New("has no name")
 	}
 
+	if strings.HasPrefix(name, "_") {
```
stgraber, line 204: @gibmat do we already have detection for duplicate names within the config?

gibmat: No, we don't but it would be easy enough to add.

stgraber: Let's do that then, that will avoid situations where someone messed up some copy/paste :)

gibmat:
@stgraber I think this should be ready; the various test network configs I have all work, so hopefully there aren't any regressions with the changes.

stgraber (approved):

PR #250, opened 2025-08-13
Expanded proxy support

### LINKED ISSUE #221: Expand proxy support by using kpx
Currently our proxy support is rather limited, we basically just have:
```
http_proxy: URL
https_proxy: URL
no_proxy: LIST OF HOSTS/IPS
```

This typically works well, but we've now run into an environment where the proxy server only allows for Kerberos authentication. The best way to handle such cases is by using a local proxy daemon.
I quickly played with kpx (https://github.com/momiji/kpx) and that seems to be working quite well.

I think we should be bundling `kpx` and set the system and application proxy values to always route through it, then only dynamically update its configuration rather than have to try to update the config of everything on the system.

We'd basically be starting it up with the system, have it listen on `127.0.0.1:3128` and only in proxy mode, keeping everything else disabled.

The proxy config should then evolve to support multiple proxy servers and then rules for what traffic to send where (or exclude from proxying).

I'm thinking something like this:
```
proxy:
  servers:
    corp-kerberos:
      host: "my-krb-proxy.corp.example.net:3128"
      auth: "kerberos"
      username: "foo"
      password: "bar"
      realm: "corp.example.net"
    corp-basic:
      host: "my-basic-proxy.corp.example.net:8080"
      auth: "basic"
      username: "foo"
      password: "bar"
    corp-anonymous:
      host: "my-anonymous-proxy.corp.example.net:8000"
      auth: "anonymous"
  rules:
    - destination: "*.hated-domain.example.net|hated-domain.example.net"
      target: "none"
    - destination: "*.my-cluster.example.net|operations-center.example.net"
      target: "direct"
    - destination: "*.example.net"
      target: "corp-basic"
    - destination: "*.example.org"
      target: "corp-anonymous"
    - destination: "*"
      target: "corp-kerberos"
```

`none` and `direct` are special values, those should not be allowed as proxy server names.
`none` causes a refusal to proxy (request will fail), `direct` sends the traffic out directly without proxying.

If no rules are provided, we'll just configure things to try the servers in order, falling back to "none".

As part of landing this change, we'll need to convert the current proxy configurations to the new format.

gibmat:
Closes #221

Also removes support for migrating legacy json state files, which was deprecated with  #186.


mkosi.images/base/mkosi.extra/usr/lib/systemd/system/kpx.service
```diff
@@ -0,0 +1,6 @@
+[Unit]
+Description=kpx proxy daemon
+
+[Service]
+ExecStart=/usr/local/bin/kpx -c /var/lib/incus-os/kpx.yaml
+Restart=on-failure
```
stgraber, line 8: Given it's running on port 3128, I don't think there's a reason to run this one as root and it may be beneficial to run it as a nobody/nogroup user or similar just to limit attack surface.

stgraber (approved):

PR #257, opened 2025-08-14
kpx: Add IPv6 support

stgraber:


gibmat:
It would be worthwhile to push this upstream, too.

stgraber:
@gibmat yeah, that's the plan, if we don't see any weird side effect.
It seems quite deliberate (used `tcp4`) so I'm not sure what the reason behind that may have been.

PR #281, opened 2025-08-21
Add applications API

stgraber:


stgraber:
@gibmat can you check that I did the state upgrade correctly?

PR #287, opened 2025-08-25
Implement recovery mode

### LINKED ISSUE #234: Implement a recovery mode
On incus-osd startup, we should look for a specific partition/disk label, say `RESCUE_DATA` or something like that.

If found, we should attempt to mount it (restrict to vfat or isofs for safety) and in there, look for either:
 - Update folder (`update`) containing a full update including a signed `update.json`. If found, proceed with validating and applying that update.
 - A signed file called `hotfix.sh` (`hotfix.sh.sig` for the signature), if found and valid, execute the script.

In both cases, validation should be done using the same key as we use for official updates.

gibmat:
Closes #234


incus-osd/cmd/incus-osd/main.go
```diff
@@ -315,6 +316,14 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error {
 		slog.WarnContext(ctx, "Booted from backup "+s.OS.Name+" image version "+s.OS.RunningRelease)
 	}
 
+	// Check for and run recovery logic if present.
+	err = recovery.CheckRunRecovery(ctx, s)
```
gibmat, line 304: For now, I've put the recovery check just before we bring the network online.

incus-osd/internal/recovery/recovery.go
```diff
@@ -0,0 +1,286 @@
...
+
+	_, err = fmt.Fprintf(rootCA, providers.LxcUpdateCA)
+	if err != nil {
+		return err
+	}
+
+	defer os.Remove(rootCA.Name())
+
+	// Verify the script's signature.
+	_, err = subprocess.RunCommandContext(ctx, "openssl", "smime", "-verify", "-CAfile", rootCA.Name(), "-inform", "PEM", "-in", filepath.Join(mountDir, "hotfix.sh.sig"), "-content", filepath.Join(mountDir, "hotfix.sh"))
```
gibmat: openssl complains that the CA certificate is self-signed here, but not when checking update files below.

incus-osd/internal/providers/structs.go
```diff
@@ -5,6 +5,21 @@ import (
 	"strconv"
 )
 
+// LxcUpdateCA is used to verify updates.
+const LxcUpdateCA = `-----BEGIN CERTIFICATE-----
```
stgraber: Nit, but should really be `LXCUpdateCA` then :)

incus-osd/internal/recovery/recovery.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		return err
+	}
+
+	slog.InfoContext(ctx, "Running hotfix script")
+
+	// Run the hotfix script.
+	_, err = subprocess.RunCommandContext(ctx, "bash", filepath.Join(mountDir, "hotfix.sh"))
```
stgraber: I'd prefer we execute the script and respect its shebang.

incus-osd/internal/recovery/recovery.go
stgraber: I vaguely remember the `-text` being important as that changes the format of the signature a bit. @breml may remember the detail on that one, but I remember having to change the way we generate the signed file slightly a while back to accommodate the different format.

incus-osd/cmd/incus-osd/main.go
stgraber: I think that makes sense, though the commit title says the opposite :)

PR #290, opened 2025-08-27
Implement update API

### LINKED ISSUE #236: Implement an update API
We should implement a system update API allowing for configuration of update frequencies and maintenance windows as well as manually triggering an update.

- GET /1.0/system/update to get the current config and state
- POST /1.0/system/update to trigger a manual update

Config should allow:
 - Configuration of update frequency
 - Opt-in to automatic OS reboot following update
 - Set of maintenance windows where updates can be downloaded and applied (to avoid application downtime). We'll want to find a syntax to easily match based on day of week and time.

gibmat:
Closes #236

### DIFF
--- a/incus-osd/api/system_update.go
+++ b/incus-osd/api/system_update.go
@@ -1,0 +1,89 @@
+package api
+
+import (
+	"time"
+)
+
+// SystemUpdate defines a struct to hold information about the system's update policy.
+type SystemUpdate struct {
+	Config SystemUpdateConfig `json:"config" yaml:"config"`
+
+	State struct {
+		LastCheck    time.Time `json:"last_check"    yaml:"last_check"`
+		UpdateStatus string    `json:"update_status" yaml:"update_status"`
+		NeedsReboot  bool      `json:"needs_reboot"  yaml:"needs_reboot"`
+	} `json:"state" yaml:"state"`
+}
+
+// SystemUpdateConfig defines a struct to hold configuration details for the update checks.
+type SystemUpdateConfig struct {
+	UpdateFrequency    time.Duration                   `json:"update_frequency"              yaml:"update_frequency"`
+	AutoReboot         bool                            `json:"auto_reboot"                   yaml:"auto_reboot"`
+	MaintenanceWindows []SystemUpdateMaintenanceWindow `json:"maintenance_windows,omitempty" yaml:"maintenance_windows,omitempty"`
+}
+
+// SystemUpdateMaintenanceWindow defines a maintenance window for when it is acceptable to check for and apply updates.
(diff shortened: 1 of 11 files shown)


incus-osd/api/system_update.go
```diff
@@ -0,0 +1,105 @@
+package api
+
+import (
+	"time"
+)
+
+// Weekday defines our own type. The time package's Weekday doesn't include any way to indicate
+// a zero-value, which we need.
+type Weekday int
```
gibmat: Not a huge fan of this, but I'm not sure if there's anything better we can do.

incus-osd/cmd/incus-osd/main.go
```diff
@@ -514,16 +522,44 @@ func updateChecker(ctx context.Context, s *state.State, t *tui.TUI, p providers.
 	}
 
 	for {
-		// Sleep at the top of each loop, except if we're performing a startup check.
+		// Sleep at the top of each loop, except if we're performing a startup or manual check.
 		if !isStartupCheck && !isUserRequested {
-			time.Sleep(6 * time.Hour)
+			timeSinceCheck := time.Since(s.System.Update.State.LastCheck)
+			if timeSinceCheck < s.System.Update.Config.UpdateFrequency {
+				time.Sleep(s.System.Update.Config.UpdateFrequency - timeSinceCheck)
```
gibmat, line 529: Because of how we're relying on `sleep`, if the update frequency is changed, it won't actually take effect until the current sleep expires.

incus-osd/api/system_update.go
```diff
@@ -0,0 +1,105 @@
...
+type SystemUpdateConfig struct {
+	UpdateFrequency    time.Duration                   `json:"update_frequency"              yaml:"update_frequency"`
+	AutoReboot         bool                            `json:"auto_reboot"                   yaml:"auto_reboot"`
+	MaintenanceWindows []SystemUpdateMaintenanceWindow `json:"maintenance_windows,omitempty" yaml:"maintenance_windows,omitempty"`
+}
+
+// SystemUpdateMaintenanceWindow defines a maintenance window for when it is acceptable to check for and apply updates.
+// StartDayOfWeek and EndDayOfWeek are optional, and if non-zero can be used to limit the migration window to certain day(s).
+// Times are assumed to be in UTC.
+type SystemUpdateMaintenanceWindow struct {
```
stgraber, line 28: The API should likely take a string for the day of week.

That would avoid two issues:
 - Whether it's 0 indexed or 1 indexed
 - What the first day of the week is (it's about a 50/50 split whether a given culture uses Sunday or Monday as the first day of the week, with some also using Saturday)

stgraber (approved):

PR #296, opened 2025-09-02
Report LVM volume group information

### LINKED ISSUE #245: Report LVM VG status in LVM service endpoint
We currently don't report errors if a VG is in a bad state, that'd be good to have here.

gibmat:
Adds reporting of each volume group and any LVM logs to the LVM service endpoint.

Closes #245


stgraber:
> Would it be useful/make sense to also report physical volumes too?

Yeah, probably makes sense to report the PVs too.

PR #297, opened 2025-09-02
Add storage wipe API endpoint

### LINKED ISSUE #278: Add "wipe device" to storage API
It might be useful to add a "wipe drive" command to the storage API. This would make it easier for the user who connects a drive but forgot to wipe it first to begin using it in a new storage pool. We'd want to ensure we don't wipe any drive that currently belongs to a storage pool.

gibmat:
Adds a way to run `blkdiscard` via the API. Some safeguards are in place: you can't wipe the boot device, any remote device, or any device that's currently part of a storage pool.

Closes #278

### DIFF
--- a/incus-osd/internal/rest/api_system_storage.go
+++ b/incus-osd/internal/rest/api_system_storage.go
@@ -91,3 +91,51 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 
 	_ = s.state.Save(r.Context())
 }
+
+func (*Server) apiSystemStorageWipe(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	switch r.Method {
+	case http.MethodPost:
+		// Wipe the specified drive.
+		if r.ContentLength <= 0 {
+			_ = response.BadRequest(errors.New("no drive specified")).Render(w)
+
+			return
+		}
+
+		b, err := io.ReadAll(r.Body)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


incus-osd/internal/rest/api_system_storage.go
```diff
@@ -91,3 +91,36 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 
 	_ = s.state.Save(r.Context())
 }
+
+func (*Server) apiSystemStorageWipe(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	switch r.Method {
+	case http.MethodDelete:
```
stgraber: I'd make that a Post as it's an action. Delete would be suitable if we were deleting an entry from a collection, but we aren't.

incus-osd/internal/rest/api_system_storage.go
```diff
@@ -91,3 +91,36 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 
 	_ = s.state.Save(r.Context())
 }
+
+func (*Server) apiSystemStorageWipe(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	switch r.Method {
+	case http.MethodPost:
```
stgraber, line 99: I think we should have a JSON struct for the action.

Currently it should just be the drive identifier, but it could eventually evolve to including an option to choose what method to use to wipe the drive.

PR #309, opened 2025-09-08
Pin systemd version

gibmat:
As reported in Debian bug #1112535, systemd-networkd in 257.8-1~deb13u1 breaks VLAN+Bridge setups, which we heavily use. For now, use pinning and the snapshot.debian.org archive to install the prior version of systemd packages.


jarrodu:
Did you hit the big?

stgraber:
No but we noticed it in the Debian bug tracker and didn't feel like taking the risk of it hitting a production deployment :)

jarrodu:
I saw it too. Its a pretty crazy issue. I am sure it will resolved soon.

-------- Original Message --------

PR #319, opened 2025-09-11
Detect proxy when returning API endpoints

gibmat:
If we receive a request with the "X-IncusOS-Proxy" header set to a non-empty string, prefix all returned API endpoints with that string.

### DIFF
--- a/incus-osd/internal/rest/api_root.go
+++ b/incus-osd/internal/rest/api_root.go
@@ -2,9 +2,27 @@ package rest
 
 import (
 	"net/http"
+	"net/url"
 
 	"github.com/lxc/incus-os/incus-osd/internal/rest/response"
 )
+
+// If the request contains a X-IncusOS-Proxy header, prepend that to
+// the API root that's returned to the user.
+func getAPIRoot(r *http.Request) string {
+	if r == nil {
+		return "/1.0"
+	}
+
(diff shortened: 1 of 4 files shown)


stgraber:
@gibmat I think we should have the header contain the prefix so if we're dealing with a situation where we need to be somewhere other than /os it would just work.

incus-osd/internal/rest/api_root.go
```diff
@@ -2,10 +2,28 @@ package rest
...
 )
 
+// If the request contains a X-Incusos-Proxy header, prepend that to
+// the API root that's returned to the user.
+func getAPIRoot(r *http.Request) string {
+	if r == nil {
+		return "/1.0"
+	}
+
+	prefix := r.Header.Get("X-Incusos-Proxy")
```
stgraber: Nit but since we already have capitalization in there, should make it `X-IncusOS-Proxy`

gibmat: Yeah, that's what I had before but then the linter in the prior CI run was complaining about it. Locally with the same version of the linter I wasn't getting that warning.

stgraber: I'd argue the linter is wrong and we should override it then :)

gibmat: And now on a different run with the same input the linter is happy. Yay for deterministic tools....

PR #332, opened 2025-09-16
Add logic to set encryption key for existing storage pool

gibmat:
This adds logic and an endpoint to provide the encryption key when importing an existing storage pool. The use case is being able to migrate an existing storage pool to a new IncusOS install, or recovery if IncusOS needs to be installed. If the encryption key had been previously saved, it can be provided and IncusOS will save it locally to allow decryption of the storage pool.

### DIFF
--- a/incus-osd/api/system_storage.go
+++ b/incus-osd/api/system_storage.go
@@ -61,3 +61,11 @@ type SystemStorageDriveSMART struct {
 type SystemStorageWipe struct {
 	ID string `json:"id" yaml:"id"`
 }
+
+// SystemStoragePoolKey defines a struct used to provide an encryption key when importing an existing pool.
+// Currently the only supported type is "zfs".
+type SystemStoragePoolKey struct {
+	Name          string `json:"name"           yaml:"name"`
+	Type          string `json:"type"           yaml:"type"`
+	EncryptionKey string `json:"encryption_key" yaml:"encryption_key"`
+}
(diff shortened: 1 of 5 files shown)


incus-osd/api/system_storage.go
```diff
@@ -61,3 +61,9 @@ type SystemStorageDriveSMART struct {
 type SystemStorageWipe struct {
 	ID string `json:"id" yaml:"id"`
 }
+
+// SystemStoragePoolKey defines a struct used to provide a pool's encryption key when importing an existing pool.
+type SystemStoragePoolKey struct {
```
stgraber, line 67: We should add a key to specify the pool type, similar to what we do in the rest of the storage API.

gibmat: Do we really need to know that though? The pool will already exist, so if we need to know any details we should be able to directly query the pool. For ZFS, all we need to do is make the raw encryption key available at the expected location.

Down the road if we add support for additional storage backends I'd expect the `PoolExists()` method to return the pool's type (zfs, lvm, etc), which would then inform the logic for properly dealing with encryption configuration.

stgraber: What happens if you have both an existing LVM VG named `foo` and you're trying to import a zpool named `foo` too?

We're not expecting pool names to be globally unique, the guarantee is on type+poolname.

stgraber:
It's going to need a quick rebase on main to pick up the linter fixes.

PR #333, opened 2025-09-17
Add logic to uninstall an application

gibmat:
Add logic to uninstall an application, optionally also removing all relevant user data as well. Extends the `/applications/{name}` endpoint to accept a `DELETE` request to trigger the removal.

### DIFF
--- a/incus-osd/internal/applications/app_operations_center.go
+++ b/incus-osd/internal/applications/app_operations_center.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"os"
 	"slices"
 	"time"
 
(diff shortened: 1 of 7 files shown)


incus-osd/internal/applications/app_operations_center.go
```diff
@@ -173,6 +174,42 @@ func (*operationsCenter) IsRunning(ctx context.Context) bool {
...
+	if err != nil {
+		return err
+	}
+
+	// Reload the systemd daemon.
+	err = systemd.ReloadDaemon(ctx)
+	if err != nil {
+		return err
+	}
+
```
stgraber, line 202: Given the repetition, should we have a `systemd.RemoveExtension` or something which does those three steps?

stgraber:
I'm not sure what the use case for this is, at least for primary applications (incus, operations-center, migration-manager) as the moment you uninstall them, you've got a bricked system.

It makes sense for additional applications like debug or the openfga application, but not so much for the primary ones.

PR #343, opened 2025-09-24
Add a system log endpoint to enable logging messages received via REST

### LINKED ISSUE #335: Add API endpoint to display message via logger
It would be useful to add an API endpoint that would accept a string and then log it via `slog` to display on screen and in the journal.

gibmat:
Closes #335

### DIFF
--- a/doc/rest-api.md
+++ b/doc/rest-api.md
@@ -25,7 +25,11 @@ are not backwards compatible.
   
     `GET`: Returns systemd journal entries, optionally filtering by unit, boot number, and
     number of return entries
+
+  * `/1.0/debug/tui`
   
+    `POST`: Send a message that should be logged by the system
+
   * `/1.0/services`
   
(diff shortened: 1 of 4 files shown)


doc/rest-api.md
```diff
@@ -42,7 +42,11 @@ are not backwards compatible.
       - `shutdown`, `poweroff`, `reboot`: Self-descriptive
       - `reset_encryption_bindings`: Force-reset TPM encryption bindings; intended only for
       use if it was required to enter a recovery passphrase to boot the system
+ 
+ * `/1.0/system/log`
```
stgraber: I think we probably should do `POST /1.0/debug/tui` instead so we don't have a weird mix of `/1.0/system/log` and `/1.0/debug/log`. This is very much an internal/debug route so I don't think it should be on /1.0/system.

PR #356, opened 2025-09-29
Add OpenFGA app

### LINKED ISSUE #354: Create OpenFGA application
We should build a supplemental application for `openfga` which should come along with a PostgreSQL database.
This will commonly be deployed alongside `operations-center` so servers can also handle authorization.

gibmat:
When initializing, the application will generate an initial random authentication token. Authentication tokens can be changed via the application configuration API.

The `openfga` service will start its HTTPS listener on port 8444.

Each time the application starts, it will grab a copy of the primary application's TLS certificate/key and use those to secure the OpenFGA API endpoints. When the primary application TLS certificate changes, just restart the OpenFGA application to pickup the new certificate.

```
bash-5.2# curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/applications/openfga | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   337  100   337    0     0   209k      0 --:--:-- --:--:-- --:--:--  329k
{
  "type": "sync",
  "status": "Success",
  "status_code": 200,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": {
    "state": {
      "initialized": true,
      "friendly_version": "1.17.1 [202606111508]",
      "version": "202606111508",
      "available_versions": [
        "202606111508"
      ]
    },
    "config": {
      "api_tokens": [
        "45DWOCMJSTFXGV4BJBON7CVOSB"
      ]
    }
  }
}
```

Closes #354

### DIFF
--- a/incus-osd/api/application_openfga.go
+++ b/incus-osd/api/application_openfga.go
@@ -1,0 +1,20 @@
+package api
+
+// ApplicationOpenFGAConfig represents additional configuration for the openfga application.
+type ApplicationOpenFGAConfig struct {
+	ApplicationConfig
+
+	APITokens []string `json:"api_tokens" yaml:"api_tokens"`
+}
+
+// ApplicationOpenFGAState represents the state of the openfga application.
+type ApplicationOpenFGAState struct {
+	ApplicationState
+}
+
+// ApplicationOpenFGA represents the state and configuration of the openfga application.
+type ApplicationOpenFGA struct {
+	State ApplicationOpenFGAState `json:"state" yaml:"state"`
+
+	Config ApplicationOpenFGAConfig `json:"config" yaml:"config"`
+}
--- a/incus-osd/internal/applications/app_openfga.go
+++ b/incus-osd/internal/applications/app_openfga.go
@@ -1,0 +1,443 @@
+package applications
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"time"
+
+	"github.com/lxc/incus/v7/shared/subprocess"
+	"go.yaml.in/yaml/v4"
+	"golang.org/x/sys/unix"
+
+	"github.com/lxc/incus-os/incus-osd/api"
+	"github.com/lxc/incus-os/incus-osd/internal/systemd"
+	"github.com/lxc/incus-os/incus-osd/internal/zfs"
+)
+
+type openfga struct {
+	common
+}
+
+type openfgaConfig struct {
+	Authn openfgaConfigAuthn `yaml:"authn"`
+	Grpc  openfgaConfigGrpc  `yaml:"grpc"`
+	HTTP  openfgaConfigHTTP  `yaml:"http"`
+}
+
+type openfgaConfigAuthn struct {
+	Method    string                      `yaml:"method"`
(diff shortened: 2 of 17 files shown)


incus-osd/api/application_openfga.go
```diff
@@ -0,0 +1,24 @@
...
+
+// ApplicationOpenFGAConfig represents additional configuration for the openfga application.
+type ApplicationOpenFGAConfig struct {
+	ApplicationConfig
+
+	APITokens   []string `json:"api_tokens"             yaml:"api_tokens"`
+	ListenAddr  string   `json:"listen_addr,omitempty"  yaml:"listen_addr,omitempty"`  // If not specified, defaults to 127.0.0.1.
+	GrpcPort    int      `json:"grpc_port,omitempty"    yaml:"grpc_port,omitempty"`    // If not specified, defaults to 8081.
+	HTTPPort    int      `json:"http_port,omitempty"    yaml:"http_port,omitempty"`    // If not specified, defaults to 8080.
+	MetricsPort int      `json:"metrics_port,omitempty" yaml:"metrics_port,omitempty"` // If not specified, defaults to 2112.
```
stgraber: Hmm, I think we need to start much simpler / opinionated for now.
I would do away with all of those, completely disable the metrics and grpc port, disable HTTP in favor of HTTPS, pick a port for it that doesn't conflict with Operations Center (I'd do likely go the easy route with 8444) and for the listen address, bind to everything.

For the certificate, because this isn't a primary application, we should ensure it only starts AFTER a primary application has started. We'll then reuse the certificate and key of the primary application for the HTTPS endpoint.

stgraber:
@gibmat tweak to default config so we can get this going with HTTPS and eliminate just about everything else for now :)

incus-osd/api/application_openfga.go
gibmat: OK, I've trimmed away those options. :)

There's not a way to disable the gRPC server, so I set things up for the HTTPS server to listen on `:8444` with the gRPC on `:8445`.

stgraber: Can you move gRPC to the loopback address then so we don't have a random gRPC API exposed externally?

stgraber:
Just so it doesn't get lost on the solved comment. I don't really want us to expose the gRPC API publicly when none of our software use it. So given it can't be disabled, we should bind it to localhost and probably bind it to a high random-ish port so it's not likely to conflict with anything the user may use (8445 seems risky in that regard, 58444 or something less so).

incus-osd/internal/applications/app_openfga.go
```diff
@@ -0,0 +1,443 @@
...
+		Grpc: openfgaConfigGrpc{
+			Addr: "localhost:55123", // Since we can't disable the gRPC service, bind to localhost only on a high port.
+			TLS: openfgaConfigTLS{
+				Enabled: true,
+				Cert:    "/var/lib/openfga/server.crt",
+				Key:     "/var/lib/openfga/server.key",
+			},
+		},
+		HTTP: openfgaConfigHTTP{
+			Addr: "0.0.0.0:8444",
```
stgraber: Does that bind IPv6 as well?
I'd have expected `:8444` to bind both IPv4 and IPv6.

gibmat: Yes, it does, but I'll change to `:8444` to make that a bit more obvious.

PR #358, opened 2025-09-30
Implement OS backup/restore

gibmat:
TODO:
* Need to properly handle some values of the new state struct to make sure they correctly reflect the restored system
* A clean way to install/uninstall a given application

### DIFF
--- a/doc/rest-api.md
+++ b/doc/rest-api.md
@@ -47,6 +47,17 @@ are not backwards compatible.
       - `reset_encryption_bindings`: Force-reset TPM encryption bindings; intended only for
       use if it was required to enter a recovery passphrase to boot the system
 
+  * `/1.0/system/backup`
+  
+    `GET`: Return a tar archive backup of the system state and configuration
+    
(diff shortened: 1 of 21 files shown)


doc/rest-api.md
```diff
@@ -59,6 +59,17 @@ are not backwards compatible.
...
+  
+    `GET`: Return a tar archive backup of the system state and configuration
+    
+    `PATCH`: Use provided tar archive to perform a partial restore of the system state and
+    configuration. Things like the "local-data" encryption key and NIC MAC addresses will not
+    be overwritten from the provided backup.
+    
+    `PUT`: Use provided tar archive to perform a complete restore of the system state and
+    configuration.
+
```
stgraber: This API feels wrong given we're really not dealing with a JSON REST endpoint.

To align better with https://github.com/lxc/incus-os/issues/362, I think we should instead have:

 - POST /1.0/system/:backup
 - POST /1.0/system/:restore

Handling of the partial case, should be done through parameters on the `:restore` endpoint, so something like `POST /1.0/system/:backup?skip=mac,local-data-encryption`

PR #359, opened 2025-10-01
Add application backup/restore and factory reset methods

gibmat:


incus-osd/internal/applications/app_incus.go
```diff
@@ -181,6 +182,55 @@ func (*incus) AddTrustedCertificate(_ context.Context, name string, cert string)
...
+	if err != nil {
+		return err
+	}
+
+	return os.RemoveAll("/var/log/incus/")
+}
+
+// GetBackup returns a tar archive backup of the application's configuration and/or state.
+func (*incus) GetBackup() ([]byte, error) {
+	return createTarArchive("/var/lib/incus/", []string{"guestapi", "shmounts", "unix.socket"})
```
gibmat: Not sure if we should be excluding anything else from the backup archive that's generated.

incus-osd/internal/rest/api_applications.go
```diff
@@ -60,3 +62,114 @@ func (s *Server) apiApplicationsEndpoint(w http.ResponseWriter, r *http.Request)
...
+		// Restore the application's backup.
+		err := app.RestoreBackup(r.Body)
+		if err != nil {
+			_ = response.BadRequest(err).Render(w)
+
+			return
+		}
+
+		// Restart the application.
+		err = app.Stop(r.Context(), "")
```
gibmat, line 156: Both when performing a factory reset and when restoring a backup, the application will be restarted. This will result in a connectivity blip of the API proxy.

stgraber: Sounds fine, not much we can do about it other than try to close the request properly before the proxy dies on us (knowing that we won't be able to report anything to the user after that point anyway).

PR #369, opened 2025-10-08
Rework REST API endpoints

### LINKED ISSUE #362: Rename API action endpoints
One pattern we started using with @breml in Operations Center is using `:ACTION` as the API route for an action. This then reduces the risk of eventually getting a conflict with potential other entries.

In our case, we should therefore start using:
 - `/1.0/debug/tui/:write-message`
 - `/1.0/system/:factory-reset`
 - `/1.0/system/:poweroff`
 - `/1.0/system/:reboot`
 - `/1.0/system/storage/:import`
 - `/1.0/system/storage/:wipe`
 - `/1.0/system/update/:refresh`

That also has the nice side-effect of making it very easy to spot all such special commands and making sure we have them covered in the CLI.

### LINKED ISSUE #364: Implement `GET /1.0/system`
We should make `GET /1.0/system` act like a collection and give us a list of endpoint.

gibmat:
* Rename endpoints to include `:ACTION`
* Implement GET /1.0/system
* Add a missing debug endpoint in returned list

Hopefully this is obvious, but this is a big breaking change in the REST APIs for IncusOS. :slightly_smiling_face: 

Closes #362, Closes #364

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -155,7 +155,7 @@ jobs:
           incus file push --quiet mkosi.output/debug.raw test-incus-os/root/updates/
           incus file push --quiet mkosi.output/incus.raw test-incus-os/root/updates/
 
-          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update -X POST
+          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update/:check -X POST
 
           sleep 3m
 
@@ -173,7 +173,7 @@ jobs:
       - name: Test EFI db and dbx updates
         run: |
           incus file push certs/efi/updates/*.auth test-incus-os/root/updates/
-          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update -X POST
+          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update/:check -X POST
 
           sleep 30s
 
@@ -193,7 +193,7 @@ jobs:
           incus file push --quiet mkosi.output/IncusOS_${RELEASE}.efi test-incus-os/root/updates/
           incus file push --quiet mkosi.output/IncusOS_${RELEASE}.usr* test-incus-os/root/updates/
 
-          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update -X POST
+          incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update/:check -X POST
 
           sleep 3m
 
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ test-applications:
 	incus file push mkosi.output/debug.raw test-incus-os/root/updates/
 	incus file push mkosi.output/incus.raw test-incus-os/root/updates/
 
-	incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update -X POST
+	incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update/:check -X POST
 
 .PHONY: test-update
 test-update:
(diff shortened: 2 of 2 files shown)


stgraber:
The last commit looks wrong with the set-configuration type actions.

/1.0/system/XYZ and /1.0/services/XYZ are proper REST endpoint. You can GET them and PUT the same struct back for update. We don't need action endpoints using POST to do that.

doc/rest-api.md
```diff
@@ -17,18 +17,20 @@ are not backwards compatible.
   
     `GET`: Returns application-specific status and/or configuration information
 
-  * `/1.0/applications/{name}/backup`
+  * `/1.0/applications/{name}/:backup`
 
     `GET`: Returns a tar archive backup for the application. If passed `complete=true`, a
```
stgraber: I'd still make that a POST as it is an action (we're making a backup), not just retrieving something that already exists.

doc/rest-api.md
```diff
@@ -17,18 +17,20 @@ are not backwards compatible.
...
-
-  * `/1.0/applications/{name}/reset`
+  * `/1.0/applications/{name}/:factory-reset`
 
     `POST`: Performs a factory reset of the application
-  
+
+  * `/1.0/applications/{name}/:restore`
+
+    `PUT`: Restore a tar archive backup for the application
```
stgraber: POST

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
+
+  * `/1.0/system/:factory-reset`
+
+    `POST`: Perform a complete factory reset of the system and immediately reboot. An
+    optional number of seed configurations may also be provided, which will be used
+    when the system starts up into its fresh state.
+
+  * `/1.0/system/:poweroff`
+
+    `PUT`: Power off the system
```
stgraber: POST

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
+
+  * `/1.0/system/:reboot`
+
+    `PUT:` Reboot the system
```
stgraber: POST

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
 
   * `/1.0/system/network`
-  
+
     `GET`: Return the current network state
-    
+
     `PATCH`: Update the current network configuration
-    
+
```
stgraber: Is PATCH something we're actually using/testing?

If not, I'd be tempted to stay away from PATCH in our API for now, too annoying to get perfectly right.

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
     status, encryption recovery keys, etc
-    
+
+  * `/1.0/system/security/:add-encryption-key`
+
     `PUT`: Add an encryption recovery passphrase
-    
+
+  * `/1.0/system/security/:delete-encryption-key`
+
```
stgraber: That looks wrong as we can just use PUT on /1.0/system/security to update the recovery key field we get through GET

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
     `DELETE`: Remove an encryption recovery passphrase
-  
+
+  * `/1.0/system/security/:tpm-rebind`
+
+    `PUT`: Force-reset TPM encryption bindings; intended only for use if it was required to enter
+    a recovery passphrase to boot the system
```
stgraber, line 101: POST

doc/rest-api.md
```diff
@@ -38,77 +40,105 @@ are not backwards compatible.
...
-  * `/1.0/system/storage/import`
-  
+  * `/1.0/system/storage/:create-pool`
+
+    `PUT`: Create a local storage pool
+
+  * `/1.0/system/storage/:delete-pool`
+
+    `DELETE`: Destroy a local storage pool
+
```
stgraber: Creation and update can be done through `PUT` by updating the pool list in the config you get from `GET`.

We could also handle the delete case that way but that's pretty dangerous given we're dealing with actual user data in this scenario, so having a `POST /1.0/system/storage/:delete-pool` to handle that seems reasonable.

stgraber:
Looking better, still some inconsistencies but we're getting there (I've only reviewed the doc not the actual implementation at this point).

doc/rest-api.md
gibmat: There was some logic to handle a PATCH update to the network configuration, but it's easy enough to take out so we don't have to worry about it.

doc/rest-api.md
gibmat: Sure, I did some refactoring to make that work like the other endpoints to accept an API struct to update the security configuration (only the encryption recovery key(s) at the moment).

doc/rest-api.md
```diff
@@ -38,77 +40,91 @@ are not backwards compatible.
...
     pools
-    
+
     `PUT`: Create or update a local storage pool
 
-  * `/1.0/system/storage/import`
-  
+  * `/1.0/system/storage/:delete-pool`
+
+    `DELETE`: Destroy a local storage pool
```
stgraber: POST :)

PR #371, opened 2025-10-08
Limit building of the Migration Manager worker image to amd64

gibmat:
The vmware vddk isn't available for arm64, so only build the image for amd64.


stgraber:
@gibmat do we need to change the spot where we include the .iso or will it deal fine with it not existing?

gibmat:
> @gibmat do we need to change the spot where we include the .iso or will it deal fine with it not existing?

That's exactly what I realized. :grin:

gibmat:
OK, should be ready now.

PR #377, opened 2025-10-09
Add provider REST endpoint

### LINKED ISSUE #363: Allow enrolling with a provider
Currently it's only possible to enroll with a provider like `operations-center` through the `provider.yaml` seed at installation time.

This poses an issue when dealing with existing systems that should then be moved under control of a given provider. Most commonly this happens when we want to add an Operations Center server running on IncusOS to be managed by itself (so we get the benefit of update control, central configuration, ...).

I believe we should:
 - Cleanup the current `/1.0/system/provider` API handler (should be under `api_system_provider.go`)
 - Add support for `PUT` on that endpoint allowing changing of provider from the default (`images`) over to an alternative one (currently only `operations-center`).
 - Doing this would then not only change the internal state struct but also call Register()
 - Register() should be changed to get an argument so it can know if it's a first-boot enrollment or something that's done later, mostly so it has an option to reject the change depending on overall system state.
 - We should also add a Deregister() function that would be called ahead of switching provider and again, allow for a provider to reject the change. In the current state of things, operations-center would always fail the Deregister() call whereas the images provider would always succeed it.

gibmat:
Closes #363

### DIFF
--- a/doc/rest-api.md
+++ b/doc/rest-api.md
@@ -107,6 +107,12 @@ are not backwards compatible.
 
     `PUT`: Apply a new network configuration
 
+  * `/1.0/system/provider`
+
+    `GET`: Returns the current system provider state
+
+    `PUT`: Apply a new system provider configuration
+
   * `/1.0/system/resources`
 
     `GET`: Returns a detailed low-level dump of the system's resources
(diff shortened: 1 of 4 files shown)


gibmat:
Did a quick rebase to fix the calls to `s.state.Save()` after the OS backup PR was merged.

PR #393, opened 2025-10-16
Add system and API tests

### LINKED ISSUE #376: Setup API and regression testing
IncusOS now has a fairly large number of API endpoints, but there's really no testing framework in place. We should introduce a suite of tests to exercise each API to ensure it works and doesn't accidentally regress.

I don't think we'd want to run the whole suite on each PR or commit, but an overnight or weekly run would probably be sufficient.

gibmat:
Add an initial set of API tests and setup a daily workflow to run them. More tests are in progress, but let's get what's ready merged and we'll add additional tests in the future.

Closes #376

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,0 +1,34 @@
+name: Daily
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+permissions:
+  contents: read
+
+jobs:
+  api-tests:
+    name: IncusOS API tests
+    strategy:
+      fail-fast: false
+    timeout-minutes: 360
+    runs-on:
+      - self-hosted
+      - cpu-4
+      - mem-8G
+      - disk-100G
+      - arch-amd64
+      - image-debian-13
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Incus
+        run: |
+          curl https://pkgs.zabbly.com/get/incus-daily | sudo sh
+          sudo chmod 666 /var/lib/incus/unix.socket
+          incus admin init --auto
+
+      - name: Run API tests
+        run: |
+          incus-osd/tests/api-tests.py
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+![Daily API tests](https://github.com/lxc/incus-os/actions/workflows/daily.yml/badge.svg)
+
 # Introduction
 Incus OS is a minimal immutable OS image dedicated to running [Incus](https://linuxcontainers.org/incus).
 It's based on [Debian](https://www.debian.org) trixie and built using [mkosi](https://github.com/systemd/mkosi).


tests/incus_test_vm/tests.py
```diff
@@ -0,0 +1,73 @@
+from inspect import getmembers, isfunction
+
```
baconYao:
```suggestion

SUFFIX_IMG = "img"
SUFFIX_ISO = "iso"

```

tests/incus_test_vm/tests.py
```diff
@@ -0,0 +1,73 @@
...
+from . import tests_incusos_api, tests_incusos_api_applications, tests_incusos_api_debug, tests_incusos_api_services, \
+    tests_incusos_api_system, tests_incusos_api_system_logging, tests_incusos_api_system_network, \
+    tests_incusos_api_system_provider, tests_incusos_api_system_resources, tests_incusos_api_system_security, \
+    tests_incusos_api_system_storage, tests_incusos_api_system_update, tests_install_external_seed, tests_install_smoke, \
+    tests_install_system_checks, tests_recovery, tests_secureboot_key_rotation, tests_seed_applications, tests_seed_install, \
+    tests_upgrade
+
```
baconYao:
```suggestion

def collect_tests(modules, image, suffix=SUFFIX_IMG):
    """Collect only functions starting with 'Test' from given modules."""
    collected = []
    for mod in modules:
        for name, fn in getmembers(mod, isfunction):
            if not name.startswith("Test"):
                continue
            collected.append([f"{name}/{suffix}", fn, image])
    return collected

```

tests/incus_test_vm/tests.py
```diff
@@ -0,0 +1,73 @@
...
+        ret.append([name + "/img", fn, image_img])
+
+    # SecureBoot key rotation tests
+    for name, fn in getmembers(tests_secureboot_key_rotation, isfunction):
+        ret.append([name + "/img", fn, image_img])
+
+    # Recovery mode tests
+    for name, fn in getmembers(tests_recovery, isfunction):
+        if "_" not in name:
+            ret.append([name + "/img", fn, image_img])
```
baconYao:
```suggestion
    # mapping table for modules and image
    mapping = [
        # Pre-install checks
        ([tests_install_system_checks], image_img, SUFFIX_IMG),
        # Smoke tests
        ([tests_install_smoke], image_img, SUFFIX_IMG),
        ([tests_install_smoke], image_iso, SUFFIX_ISO),
        # Seed-related
        ([tests_seed_applications], image_img, SUFFIX_IMG),
        ([tests_seed_install], image_img, SUFFIX_IMG),
        ([tests_install_external_seed], image_img, SUFFIX_IMG),
        # Upgrade tests
        ([tests_upgrade], old_image_img, SUFFIX_IMG),
        # IncusOS API tests
        (
            [
                tests_incusos_api,
                tests_incusos_api_applications,
                tests_incusos_api_debug,
                tests_incusos_api_services,
                tests_incusos_api_system,
                tests_incusos_api_system_logging,
                tests_incusos_api_system_network,
                tests_incusos_api_system_provider,
                tests_incusos_api_system_resources,
                tests_incusos_api_system_security,
                tests_incusos_api_system_storage,
                tests_incusos_api_system_update,
            ],
            image_img,
            SUFFIX_IMG,
        ),
        # SecureBoot key rotation tests
        ([tests_secureboot_key_rotation], image_img, SUFFIX_IMG),
        # Recovery tests
        ([tests_recovery], image_img, SUFFIX_IMG),
    ]

    for modules, image, suffix in mapping:
        ret.extend(collect_tests(modules, image, suffix))
```

baconYao (commented):
IMHO, after reviewing this piece of code, I performed the following refactoring:

1. Introduced a single mapping structure to reduce repetitive for-loops, improving readability and maintainability.

1. Added a helper function collect_tests to retrieve all functions starting with Test from the specified modules.

tests/incus_test_vm/tests.py
gibmat: Thanks for the suggestions; I've incorporated a modified version which will help to improve readability/maintainability as more tests are added.

gibmat:
@stgraber is there any way to test the new "daily" workflow before merging?

stgraber:
> @stgraber is there any way to test the new "daily" workflow before merging?

Remove the daily trigger and have it run as part of a PR push too (same as the testing one)?

stgraber:
@gibmat may want to run the python script unbuffered so we can actually get the output as it runs :)

.github/workflows/daily.yml
```diff
@@ -0,0 +1,40 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  api-tests:
+    name: IncusOS API tests
+    strategy:
+      fail-fast: false
+    timeout-minutes: 60
```
stgraber: I suspect that may not be long enough

gibmat: For comparison, I can run 4 test VMs in parallel on my machine, and the total run is about 19 minutes. The first attempt at running 4 concurrent tests in the runner failed, so I want to see how long it takes when only running one test at a time.

gibmat: Only 23 of the 33 tests were able to complete in two hours, which seems _really_ slow to me.

stgraber: Yeah, though for a daily test, we don't care too much about slow.
It can pick one of two servers with pretty vastly different CPU performance, though similar storage performance in both cases.

gibmat: I bumped its timeout to 6 hours, which should hopefully be long enough. :slightly_smiling_face: 

I also doubled the wait time for log message checks, which should hopefully fix the one failure in the latest run. (It passes fine for me locally.)

stgraber (approved):

Issue #404, opened 2025-10-20
Generate and roll out 2026 signing key


stgraber:
We need to prepare the 2026 signing key and seed it to existing systems so they're ready to go when we cut over to the new signing key at the beginning of the year.

stgraber:
The new signing keys have been generated offline and the public half of them has been uploaded to https://images.linuxcontainers.org/os/keys/

I also sent the UEFI transition bits (.auth/.esl) over to @gibmat so he can prepare the actual roll out to IncusOS users.

stgraber:
As part of this, I've also added a script in the encrypted media we use for the keys which will generate all needed artifacts for a given year. That will make 2027 extremely easy to put together.

PR #410, opened 2025-10-21
Refactor SecureBoot code

gibmat:
- Split up the monolith secureboot.go file into smaller ones
- Move most of the logic for actually applying a SecureBoot update out of incus-osd and into the secureboot package
- Add a debug API endpoint that can be used to manually test a SecureBoot update archive.

Partially covers #404

### DIFF
--- a/doc/rest-api.md
+++ b/doc/rest-api.md
@@ -42,6 +42,10 @@ are not backwards compatible.
     `GET`: Returns systemd journal entries, optionally filtering by unit, boot number, and
     number of return entries
 
+  * `/1.0/debug/secureboot/:update`
+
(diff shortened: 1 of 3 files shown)


doc/rest-api.md
```diff
@@ -42,6 +42,10 @@ are not backwards compatible.
     `GET`: Returns systemd journal entries, optionally filtering by unit, boot number, and
     number of return entries
 
+  * `/1.0/debug/:secure-boot-update`
```
stgraber: Maybe `/1.0/debug/secureboot/:update` instead?
It's the debug API so not super important, but we did use sub-paths elsewhere.

PR #417, opened 2025-10-22
Tweak system update API and make update channel configurable

### LINKED ISSUE #392: Extend update API to allow selection of update channel
We currently have the providers pick whatever the latest build is regardless of channel.
It would be good to have a configuration option in /1.0/system/update to select the desired channel (defaulting to `stable`). This would then be used by the providers (images, operations-center) to filter the candidate images.

### LINKED ISSUE #403: Hash all downloaded update files and check against expected hash from index
This is effectively a security fix as currently a compromised update server could use a valid image index but altered update files, which we'd then happily download and unpack.

It's not a critical issue because the content of the updates is signed too (dm-verity) so the kernel will refuse to load them, but it could still be used to trick a system into a DoS by downloading bad data which would then jam an update.

stgraber:
Closes #392
Closes #403

### DIFF
--- a/incus-osd/cmd/incus-osd/main.go
+++ b/incus-osd/cmd/incus-osd/main.go
@@ -83,11 +83,6 @@ func main() {
 	// Perform the install check here, so we don't render the TUI footer during install.
 	s.ShouldPerformInstall = install.ShouldPerformInstall()
 
-	// If the update frequency is set to less than five minutes, reset to a default of six hours.
-	if s.System.Update.Config.CheckFrequency.Minutes() < 5 {
-		s.System.Update.Config.CheckFrequency = 6 * time.Hour
-	}
-
 	// Clear the reboot flag on startup.
 	s.System.Update.State.NeedsReboot = false
 
--- a/incus-osd/internal/state/file.go
+++ b/incus-osd/internal/state/file.go
@@ -2,6 +2,7 @@ package state
 
 import (
 	"os"
+	"time"
 
 	"github.com/lxc/incus-os/incus-osd/api"
 )
@@ -64,6 +65,9 @@ func (s *State) Save() error {
 func (s *State) initialize() error {
 	// Use the default update channel.
 	s.System.Update.Config.Channel = "stable"
+
+	// Set the initial update frequency to 6 hours.
+	s.System.Update.Config.CheckFrequency = 6 * time.Hour
 
 	return nil
 }


stgraber:
@gibmat can you review this one?

There are a few potentially tricky pieces:
 - State version bump for the key renames and addition of the channel name
 - Refactoring of image provider logic
 - Refactoring of operations-center provider logic
 - Addition of SHA256 checking in provider logic

incus-osd/cmd/incus-osd/main.go
```diff
@@ -84,8 +84,8 @@ func main() {
 	s.ShouldPerformInstall = install.ShouldPerformInstall()
 
 	// If the update frequency is set to less than five minutes, reset to a default of six hours.
-	if s.System.Update.Config.UpdateFrequency.Minutes() < 5 {
-		s.System.Update.Config.UpdateFrequency = 6 * time.Hour
+	if s.System.Update.Config.CheckFrequency.Minutes() < 5 {
+		s.System.Update.Config.CheckFrequency = 6 * time.Hour
```
gibmat: Let's move this initialization into the new state initialize method.

incus-osd/internal/providers/provider_images.go
```diff
@@ -87,72 +86,64 @@ func (*images) GetSecureBootCertUpdate(ctx context.Context, _ string) (SecureBoo
...
 		return nil, err
 	}
 
-	// Verify the list of returned assets contains a "<name>.raw.gz" file, otherwise
-	// we shouldn't return an application update.
-	foundUpdateFile := false
-
-	for _, asset := range p.releaseAssets {
-		fileName := filepath.Base(asset)
+	// Check that an OS update is included.
```
gibmat: Minor typo -- we're checking an application

incus-osd/internal/providers/provider_operations_center.go
```diff
@@ -158,72 +178,64 @@ func (*operationsCenter) GetSecureBootCertUpdate(_ context.Context, _ string) (S
...
-	err := p.checkRelease(ctx)
+	latestUpdate, err := p.checkRelease(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// Verify the list of returned assets contains a "<name>.raw.gz" file, otherwise
-	// we shouldn't return an application update.
-	foundUpdateFile := false
+	// Check that an OS update is included.
```
gibmat: Also typo here

incus-osd/internal/providers/provider_images.go
```diff
@@ -86,72 +85,64 @@ func (*images) GetSecureBootCertUpdate(ctx context.Context, _ string) (SecureBoo
 	return &update, nil
 }
 
-func (p *images) GetOSUpdate(ctx context.Context, osName string) (OSUpdate, error) {
```
gibmat, line 89: It looks like this PR is eliminating the need to pass in the OS name for various functions in the images and operations-center providers. The local provider still uses it, but maybe it would be cleaner to remove `osName` from the methods and use a hard-coded OS name for local?

gibmat (changes requested):
Looks good overall, just a couple cleanup suggestions.

incus-osd/internal/providers/provider_images.go
stgraber: Yeah, I was looking into that. I think we'll be able to nuke it from everywhere once we also refactor the SecureBoot update logic, so was going to do that when I do that particular code update.

stgraber:
@gibmat pushed an update that should address all review comments

PR #446, opened 2025-10-28
Make support of remote in command line optional

breml:
Currently, Operations Center cli does not support passing the remote as command line argument. One has to explicitly switch to the desired remote and then perform the respective commands.

Without the changes in this PR, the description of the `admin os` commands do look confusing for an Operations Center CLI user, since it mentions the remote argument, which is in fact without any effect.

IMPORTANT: the way, this is currently implemented, this would "break" the implementation in `incus` cli, since the default (zero) value for `SupportsRemote` is `false` and therefore it will remove the mention of *remote* in `incus` cli.

I prefer to have these flags with "positive" names, but if we can not do the "breaking" change, then we would need to rename `SupportsRemote` to something like `WithoutRemote`. Let me know, what you prefer.

### DIFF
--- a/incus-osd/cli/generic.go
+++ b/incus-osd/cli/generic.go
@@ -168,7 +168,13 @@ func (c *cmdGenericList) command() *cobra.Command {
 
 func (c *cmdGenericList) run(cmd *cobra.Command, args []string) error {
 	// Quick checks.
-	exit, err := cli.CheckArgs(cmd, args, 0, 1)
+	maxArgs := 0
+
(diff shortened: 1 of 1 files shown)


stgraber:
@breml looks good but there's something missing I think.

We currently allow one optional argument with `CheckArgs(0, 1)` to the generated `List` commands as well as to the top level `Show` command. That argument is only allowed because of the remote, if remotes aren't enabled, all List commands and the top-level command should not allow any argument `CheckArgs(0, 0)`.

PR #450, opened 2025-10-28
Make CheckFrequency a string and allow disabling checks

stgraber:


incus-osd/internal/state/state_test.go
```diff
@@ -73,6 +75,7 @@ System.Provider.Config.Name: local
 System.Provider.Config.Config[multiline_value]: first\nsecond\nthird
 System.Security.Config.EncryptionRecoveryKeys[0]: ebbbibiu-ltgjfuhk-gvutdrvu-hijhvfje-gvlrgrfv-ndekdtdh-ghteuklj-ldedfifb
 System.Security.State.EncryptionRecoveryKeysRetrieved: true
+System.Update.Config.UpdateFrequency: 21600000000000
```
gibmat, line 78: `System.Update.Config.UpdateFrequency` didn't exist prior to v3, so no need to include it in earlier gold versions.

incus-osd/internal/state/upgrades.go
```diff
@@ -121,6 +122,31 @@ System.Network.Config.Proxy.Rules[%d].Target: direct
...
+			frequency, err := strconv.ParseInt(parts[1], 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("bad frequency: %d", frequency)
+			}
+
+			// Convert to a number of hours.
+			lines[i] = fmt.Sprintf("System.Update.Config.CheckFrequency: %dh", frequency/3600000000000)
+		}
+
+		lines = append(lines, "System.Update.Config.Channel: stable")
```
gibmat: Accidental copy-n-paste leftover

incus-osd/internal/state/upgrades.go
```diff
@@ -121,6 +122,31 @@ System.Network.Config.Proxy.Rules[%d].Target: direct
...
+				return nil, fmt.Errorf("malformed line '%s'", lines[i])
+			}
+
+			frequency, err := strconv.ParseInt(parts[1], 10, 64)
+			if err != nil {
+				return nil, fmt.Errorf("bad frequency: %d", frequency)
+			}
+
+			// Convert to a number of hours.
+			lines[i] = fmt.Sprintf("System.Update.Config.CheckFrequency: %dh", frequency/3600000000000)
```
gibmat: Would it be better to cast `frequency` to a `time.Duration`, then just `.String()` it to get an accurate representation of whatever value it held?

stgraber: Ah yeah, I'll do that.

incus-osd/internal/state/state_test.go
stgraber: I don't think I can fix that one.

The initial UpdateFrequency was introduced through hard-coded Go logic rather than through a state upgrade function, as a result, it doesn't get populated in the state when upgrading from V2 to V3. As the test logic validates that each of the state versions ends up being identical following upgrade, only introducing it in the V3 golden state doesn't work.

My two main options were:

1) Pretend it always existed
2) Modify the V3 state upgrade to introduce it in there

Of the two, I went with 1) as that's only touching test code. I don't feel comfortable modifying an old state update function as we then have diverging results between actual users who ran through it and what we're testing.

stgraber:
@gibmat ready for another review (I added the `never` option)

gibmat (approved):

stgraber:
Looks like there's an actual problem when running this change. I'll have to test in a VM.

PR #452, opened 2025-10-28
Cleanup providers API

gibmat:
* Remove the `osName` parameter from methods
* Be consistent with naming of targetPath parameter

Based off of #451, which should be merged first.


stgraber:
@gibmat can you give this one a quick rebase?

PR #454, opened 2025-10-29
Generate manifests

### LINKED ISSUE #379: Generate useful manifests and changelogs
mkosi provides a JSON manifest for each image that contains packages.

That's good but not quite good enough for us because:
 - No manifest is generated for images that don't install additional packages
 - Images that do install additional packages also get every package from their parent image included in their manifest
 - There is no tracking for non-package artifacts
 - There is no tracking for the build tooling used as part of the image build

I propose we effectively consume and extend the mkosi manifest so we have a JSON manifest for each image which contains:
 - All packages for that particular image (excludes anything inherited from a parent image)
 - All additional software included in the image (typically binaries we build and directly include)
 - For additional software package, the manifest should have a `build_packages` section with package entries for everything that was used to build the artifact but isn't itself part of the image (Go compiler, Javascript compiler, ...)

That will then give us a clean JSON manifest for each image we produce.

Then once we have that, we'll need a few changes to `image-publisher`:
 - Have it pull in the manifests and include them as files in the image
 - Generate human-readable changelogs by diffing the manifest of the image being imported with those of previously published images

That human readable changelog should probably look something like:
```
base:
 - Added linux-image-6.16.11-zabbly+
 - Removed linux-image-6.16.10-zabbly+
 - Updated linux-libc-dev from version XYZ to version XYZ

incus:
 - Updated incus from version XYZ to version XYZ
 - Updated incus-base from version XYZ to version XYZ
 - Updated incus-client from version XYZ to version XYZ
 - Updated incus-ui-canonical from version XYZ to version XYZ

migration-manager:
 - Updated migration-manager from version XYZ to version XYZ

operations-center:
 - Updated operations-center from version XYZ to version XYZ
```

If possible, we can keep it as valid YAML (as above) which would allow text and web UI to do some minimal re-formatting.

We'll then want those generated changelogs included as files in the image too so they can get be made available to image server clients (operations center and IncusOS itself). Once we resume making Github releases, we'd also use that text as the changelog for the release.

gibmat:
Three example changlogs: two incremental "testing" channel releases and a cumulative "stable" channel release:

```
current_version: "202511012357"
prior_version: "202511012356"
channel: testing
components:
    IncusOS:
        added:
            - bizbaz version 1.0.0
        updated:
            - adduser version 3.152 to version 3.153
            - ovn-common version 1:25.03.0-amd64-debian13-202503090556 to version 1:25.03.0-amd64-debian13-202503090556+b1
            - kpx version 1.12.2 to version 1.13.0
        removed:
            - libpolkit-gobject-1-0 version 126-2
    migration-manager:
        updated:
            - migration-manager version 0730305b74ca9005f9217c1c424a54928319a6e9 to version dcbee63dcccd37e35947fec1826a12bc5487ead6
```

```
current_version: "202511012358"
prior_version: "202511012357"
channel: testing
components:
    IncusOS:
        updated:
            - libkeyutils1 version 1.6.3-6 to version 1.6.3-7
    incus:
        updated:
            - incus version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-base version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-client version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-ui-canonical version 1:6.18-debian13-202510310310 to version 1:6.19
```

```
current_version: "202511012358"
prior_version: "202511012356"
channel: stable
components:
    IncusOS:
        added:
            - bizbaz version 1.0.0
        updated:
            - adduser version 3.152 to version 3.153
            - libkeyutils1 version 1.6.3-6 to version 1.6.3-7
            - ovn-common version 1:25.03.0-amd64-debian13-202503090556 to version 1:25.03.0-amd64-debian13-202503090556+b1
            - kpx version 1.12.2 to version 1.13.0
        removed:
            - libpolkit-gobject-1-0 version 126-2
    incus:
        updated:
            - incus version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-base version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-client version 1:6.18-debian13-202510310310 to version 1:6.19
            - incus-ui-canonical version 1:6.18-debian13-202510310310 to version 1:6.19
    migration-manager:
        updated:
            - migration-manager version 0730305b74ca9005f9217c1c424a54928319a6e9 to version dcbee63dcccd37e35947fec1826a12bc5487ead6
```

Closes #379

### DIFF
--- a/incus-osd/cmd/image-publisher/main_sync.go
+++ b/incus-osd/cmd/image-publisher/main_sync.go
@@ -181,6 +181,12 @@ func (c *cmdSync) run(cmd *cobra.Command, args []string) error {
 		})
 	}
 
+	// Generate changelog.
+	err = generateChangelog(&metaUpdate, metaUpdate.Channels[0], filepath.Join(targetPath, releaseName))
+	if err != nil {
+		return err
+	}
+
 	// Write the update metadata.
 	wr, err := os.Create(filepath.Join(targetPath, releaseName, "update.json")) //nolint:gosec
 	if err != nil {
@@ -296,6 +302,21 @@ func (*cmdSync) downloadImage(ctx context.Context, archName string, releaseURL *
 		case strings.Contains(assetName, ".usr-x86-64."), strings.Contains(assetName, ".usr-arm64."):
 			assetComponent = apiupdate.UpdateFileComponentOS
 			assetType = apiupdate.UpdateFileTypeUpdateUsr
+		case strings.HasSuffix(assetName, "debug.manifest.json"):
+			assetComponent = apiupdate.UpdateFileComponentDebug
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, "incus.manifest.json"):
+			assetComponent = apiupdate.UpdateFileComponentIncus
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, "migration-manager.manifest.json"):
+			assetComponent = apiupdate.UpdateFileComponentMigrationManager
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, "operations-center.manifest.json"):
+			assetComponent = apiupdate.UpdateFileComponentOperationsCenter
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, ".manifest.json"):
+			assetComponent = apiupdate.UpdateFileComponentOS
+			assetType = apiupdate.UpdateFileTypeImageManifest
 		default:
 			continue
 		}
--- a/incus-osd/cmd/image-publisher/changelog.go
+++ b/incus-osd/cmd/image-publisher/changelog.go
@@ -1,0 +1,146 @@
+package main
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	apiupdate "github.com/lxc/incus-os/incus-osd/api/images"
+	"github.com/lxc/incus-os/incus-osd/internal/manifests"
+)
+
+func generateChangelog(metaUpdate *apiupdate.Update, channel string, targetPath string) error {
+	slog.Info("Preparing changelog", "version", metaUpdate.Version, "channel", channel)
+
+	var metaIndex apiupdate.Index
+
+	// Get information about currently published releases.
+	// #nosec G304
+	metaFile, err := os.Open(filepath.Join(targetPath, "../", "index.json"))
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	defer func() { _ = metaFile.Close() }()
+
+	if err == nil {
+		err := json.NewDecoder(metaFile).Decode(&metaIndex)
+		if err != nil {
+			return err
+		}
+	}
+
+	priorVersion := ""
+
+	// Get the prior release, if any. Assumes updates in the index are sorted in reverse order.
+	for _, update := range metaIndex.Updates {
+		if slices.Contains(update.Channels, channel) {
+			priorVersion = update.Version
+
+			break
+		}
+	}
+
+	// Create a map of changelogs, one for each architecture.
+	changelogs := make(map[apiupdate.UpdateFileArchitecture]apiupdate.Changelog)
+
+	// For each manifest, generate its changelog entries.
+	for _, f := range metaUpdate.Files {
+		if f.Type == apiupdate.UpdateFileTypeImageManifest { //nolint:nestif
+			parts := strings.Split(f.Filename, "/")
+			archName := apiupdate.UpdateFileArchitecture(parts[0])
+			filename := parts[1]
+
+			_, exists := changelogs[archName]
+			if !exists {
+				changelogs[archName] = apiupdate.Changelog{
+					CurrnetVersion: metaUpdate.Version,
+					PriorVersion:   priorVersion,
+					Channel:        channel,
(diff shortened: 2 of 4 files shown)


app-build/build-applications.py
```diff
@@ -0,0 +1,297 @@
...
+if ARCH == "x86_64":
+    ARCH = "amd64"
+elif ARCH == "aarch64":
+    ARCH = "arm64"
+else:
+    raise("Unsupported architecture " + ARCH)
+
+applications = {
+    "kpx": {
+        "version": "1.12.2",
```
stgraber: @gibmat Can we get a `make` target which updates all the applications to the latest tag (when not using a branch like `main`)?

Basically looking for an equivalent of https://github.com/zabbly/incus/blob/daily/deps/refresh

incus-osd/cmd/generate-manifests/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/lxc/incus-os/incus-osd/internal/manifests"
+)
+
+func main() {
+	if len(os.Args) != 4 {
+		slog.Error("Usage: " + os.Args[0] + " <mkosi.output dir> <app-build dir> <output dir>")
+		os.Exit(1)
+	}
+
+	m, err := manifests.ReadManifests(os.Args[1], []string{"base", "debug", "incus", "migration-manager", "operations-center"})
```
stgraber: If we made the tool just ask for the `repository` path, it could derive:
 - output dir
 - app build dir
 - output dir

And then it could iterate over `mkosi.images/` instead of hardcoding those names.

incus-osd/api/images/update_file_type.go
```diff
@@ -41,6 +44,7 @@ var UpdateFileTypes = map[UpdateFileType]struct{}{
 	UpdateFileTypeImageRaw:                 {},
 	UpdateFileTypeImageISO:                 {},
 	UpdateFileTypeImageManifest:            {},
+	UpdateFileTypeChangelog:                {},
```
stgraber, line 47: I'd make the changes to incus-osd/api its own commit.

incus-osd/cmd/image-publisher/main_sync.go
```diff
@@ -296,6 +302,21 @@ func (*cmdSync) downloadImage(ctx context.Context, archName string, releaseURL *
...
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, "migration-manager.json"):
+			assetComponent = apiupdate.UpdateFileComponentMigrationManager
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, "operations-center.json"):
+			assetComponent = apiupdate.UpdateFileComponentOperationsCenter
+			assetType = apiupdate.UpdateFileTypeImageManifest
+		case strings.HasSuffix(assetName, ".json"):
+			assetComponent = apiupdate.UpdateFileComponentOS
+			assetType = apiupdate.UpdateFileTypeImageManifest
```
stgraber: I think we should name them `debug.manifest.json` just in case we ever have another per-application JSON blob.

stgraber:
Looking good!
Just a few nits which hopefully won't take too long to sort out!

incus-osd/cmd/generate-manifests/main.go
gibmat: At the moment mkosi doesn't generate any manifest for Operations Center, so just iterating over `*.manifest` in the mkosi.output/ directory would miss that. Would you prefer the approach of iterating the directory, then appending operations-center?

stgraber:
Oh and one more thing, in the changelog, we should probably replace `IncusOS_202511012357` with just `IncusOS` as having a versioned entry in there is going to make parsing pretty frustrating, also since the entry is a different name every time, it shouldn't be possible for it to have any changes as it stands :)

app-build/build-applications.py
gibmat: Yeah, I modified things so all the application information is a separate json file, which then makes it easy to read in the data, check for updates, and write the content back out.

gibmat:
> Oh and one more thing, in the changelog, we should probably replace `IncusOS_202511012357` with just `IncusOS` as having a versioned entry in there is going to make parsing pretty frustrating, also since the entry is a different name every time, it shouldn't be possible for it to have any changes as it stands :)

I can trim the version string from any of the components when listed in the changelog. There's actually logic in place to correctly get the prior manifest even if it has a different version, which should hopefully handle whenever we end up publishing versioned application layers.

incus-osd/cmd/generate-manifests/main.go
stgraber: I never said to iterate over mkosi.output/*.manifest, I said to iterate over the content of mkosi.images to get the list of images we're expecting.

stgraber:
```
stgraber@castiana:~/Code/lxc/incus-os (git:stgraber/main)$ ls -1 mkosi.images/
base
debug
incus
migration-manager
operations-center
```

That gives us the same list of images as you have hardcoded.

stgraber:
> > Oh and one more thing, in the changelog, we should probably replace `IncusOS_202511012357` with just `IncusOS` as having a versioned entry in there is going to make parsing pretty frustrating, also since the entry is a different name every time, it shouldn't be possible for it to have any changes as it stands :)
> 
> I can trim the version string from any of the components when listed in the changelog. There's actually logic in place to correctly get the prior manifest even if it has a different version, which should hopefully handle whenever we end up publishing versioned application layers.

Yeah, the diff looks good above, but I think we should still do the stripping as someone writing a parser shouldn't need to deal with a full release string in there as part of the component name.

incus-osd/cmd/generate-manifests/main.go
gibmat: Duh :flushed:

gibmat:
@stgraber I think all your comments have been addressed

Issue #388, opened 2025-10-14
Documentation structure and initial topics


stgraber:
Now that we have a basic documentation generator (Sphinx) and some initial doc content thanks to @gibmat, we can work on it some more and get it sufficiently complete by the time we announce IncusOS as GA.

I've got the following topics in mind, but would welcome feedback on what I may be missing!

 - Introduction
 - Getting started
   - System requirements
   - Getting an image
   - Passing in seed data for installation
   - Installing IncusOS
     - On bare metal (include example of SecureBoot key enrollment through Setup Mode or directly with DER from install media)
     - In an Incus VM (using Setup Mode)
     - In a VMWare VM (requires uploading the .der files + custom config)
     - In a Proxmox VM (steps TBD)
     - In a libivrt VM (steps TBD)
   - Accessing the system remotely
   - Configuring the system
   - Applying updates
 - Reference
   - Applications
     - Incus
     - Migration Manager
     - Operations Center
   - Services
     - Ceph
     - iSCSI
     - LVM
     - Multipath
     - NVMe
     - OVN
     - USBIP
   - System configuration
     - Logging
     - Network
     - Resources
     - Security
     - Storage
     - Update
   - API (swagger)
   - Security
 - Contributing
 - Support

stgraber:
Instructions for IncusOS on VMWare:

- Get an install ISO from https://incusos-customizer.linuxcontainers.org
- Get `secureboot-KEK-R1.der` and `secureboot-2025-R1.der` from https://images.linuxcontainers.org/os/keys/
- Upload the ISO to a VMware datastore
- Create a new VM using:
  - Create new virtual machine
  - Use Linux 6.x or later (64-bit)
  - Hardware configuration
    - 2 CPU
    - 4GB of RAM
    - 50GB of HDD
    - Add NVME controller and set HDD to NVME
    - Remove SCSI controller
    - Select ISO for CD-ROM drive and connect it
    - Add a TPM module
    - Enable Secure Boot in Boot Options
    - Set the following advanced config keys:
      - `uefi.secureBoot.kekDefault.file0` to `secureboot-KEK-R1.der`
      - `uefi.secureBoot.dbDefault.file0` to `secureboot-2025-R1.der`
  - Upload `secureboot-KEK-R1.der` and `secureboot-2025-R1.der` to the VM's folder in the datastore
  - Make sure that no .nvram files are present in the VM's folder in the datastore
  - Start the VM and let it install IncusOS
[...]

PR #480, opened 2025-11-05
Add system configuration docs

gibmat:
Partially addresses #388.


stgraber:
MACs probably should go in word list, the others should be escaped

PR #489, opened 2025-11-06
Add service docs

gibmat:
Partially addresses #388.


doc/reference/services/nvme.md
```diff
@@ -1 +1,13 @@
 # NVMe
+
+The NVMe service allows connecting a remote NVMe storage device over fibre channel or TCP.
```
stgraber, line 3: Oh, I never noticed that Fibre Channel used the British spelling :)

PR #490, opened 2025-11-06
Add Proxmox installation docs

gibmat:
Partially addresses #388.


stgraber:
Ah nice, you had access to a Proxmox deployment?

gibmat:
> Ah nice, you had access to a Proxmox deployment?

It was easy enough to spin up in an Incus VM. So I had IncusOS nested in Proxmox nested in Incus. :laughing:

PR #503, opened 2025-11-07
Add libvirt install docs

### LINKED ISSUE #498: Cannot get device ID for VirtIO-block backed install drive
When installing IncusOS in a virtual machine with a VirtIO-block backed install drive, first boot fails with the error "unable to determine device ID for /dev/disk/by-partlabel/local-data". Switching the drive to use VirtIO-scsi backing works fine.

It looks like VirtIO-block devices don't get any symlinks under `/dev/disk/by-id/`, although there are symlinks under `/dev/disk/by-uuid/`.

gibmat:
Closes #498

### DIFF
--- a/doc/getting-started/installation.md
+++ b/doc/getting-started/installation.md
@@ -9,11 +9,16 @@ should do fine. That said we recommend using generic storage and network
 adapters whenever possible, with NVMe, VirtIO or Intel virtual devices
 usually being preferred.
 
+```{note}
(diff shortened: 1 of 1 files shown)


doc/getting-started/installation.md
```diff
@@ -9,11 +9,16 @@ should do fine. That said we recommend using generic storage and network
 adapters whenever possible, with NVMe, VirtIO or Intel virtual devices
 usually being preferred.
 
+```{note}
+For virtual machines, storage drives should be configured to use the `VirtIO-sata` or `VirtIO-scsi` driver. Using `VirtIO-blk` does not work as the resulting drives will not appear to IncusOS in the same way as physical drives do.
```
stgraber: What's virtio-sata? You mean the virtual Intel SATA controller?

PR #505, opened 2025-11-07
 Enforce IP address have a CIDR mask when assigning or setting routes

### LINKED ISSUE #500: Provide an example of static network configuration and fix our address validation
@stephendotcarter ran into a situation where he had set an `addresses` entry without a CIDR mask.
This apparently got accepted but led to a broken config.

We should make sure that the documentation includes a full static IP example AND that we don't allow CDIR-less IP configuration.

gibmat:
Closes #500

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -2,6 +2,7 @@ backend
 CDN
 CDROM
 Ceph
+CIDR
 CPUs
 customizations
 customizer


doc/reference/system/network.md
```diff
@@ -47,6 +47,18 @@ Configure two network interfaces, one with IPv4 and the other with IPv6:
...
+Configure a network interface with two static IP addresses. When specifying a static IP address, it must include a CIDR mask.
+
+```
+{
+    "interfaces": [
+        {"name": "enp5s0",
+         "hwaddr": "enp5s0",
+         "addresses": ["10.234.136.100/24", "fd42:3cfb:8972:3990::100/64"]}
+    ]
+}
```
stgraber, line 66: This won't work well without providing a default gateway and DNS too, can we include a full example?

PR #508, opened 2025-11-07
Add support for extra arguments in cli package

### LINKED ISSUE #479: cli doesn't support skip options when performing system restore
When restoring a system backup via the cli, I don't think there's a way to specify any of the "skip" options (currently encryption-recovery-keys, local-data-encryption-key, or network-macs).

gibmat:
Closes #479


gibmat:
@stgraber what do you think of this approach for adding extra command-specific arguments? I modeled it off of what's currently done in `cli_debug.go`.

stgraber:
Looks pretty clean to me!

PR #512, opened 2025-11-07
 Prevent systemd from suspending the system when laptop lid is closed

### LINKED ISSUE #514: Disable all logind suspend features
A closed lid or pressed power button should not trigger a system suspend on IncusOS.

That mostly affects those running it on laptops, but fixing this would allow for the lid to get closed without the machine going to sleep.

gibmat:
Closes #514


stgraber:
Funny, I had just filed you an issue for this one ;)

PR #518, opened 2025-11-08
Clarify supported and unsupported platforms

gibmat:
Currently if you just take a glance it looks like there's a huge list of unsupported install platforms. :slightly_smiling_face:


stgraber:
Oops, I forgot to check the exact rendering, the ToC showing up in-line below definitely isn't ideal :)

PR #519, opened 2025-11-08
doc/getting-started/install/incus: Tweak wording

stgraber:
Trying to make the Secure Boot config slightly less confusing.


gibmat:
Do you want to tweak similar wording on line 51?

stgraber:
@gibmat I found the line 51 wording to be sufficiently clear here. It makes it clear that we want all keys gone so we can enroll our own. It doesn't mention MSFT but that's fine.

Issue #436, opened 2025-10-24
Add linstor support


victoitor:
Add support for listor distributed storage service/application to be configured in Incus OS.

Incus has support for linstor as a distributed storage solution and it has been mentioned that incus os will support it. Couldn't find any mention in the code, issues or PRs, so I'll just add this here.

stgraber:
Yeah, I've been thinking about this one. It's a bit trickier than what we've done with Ceph and TrueNAS so far as Linstor requires both a custom kernel module (which will need to go in the base image) and a running service (the satellite code).

The satellite code would go into the Incus application similar to what we did with Ceph, except that it would ship an actual service.

We'll also have to figure out a suitable way to build both the kernel module and the satellite code. On the kernel side, we'll need to pull something pretty recent as we track the stable kernel releases pretty closely. On the satellite, I have no idea what's involved with building and running it. Most folks use the Launchpad PPA for that, but we're not based on Ubuntu and the PPA has a tendency to pick pre-release code too (release-candidates) which isn't something we'd want to pull into IncusOS.

victoitor:
Linstor does have a public repository for [trixie](https://packages.linbit.com/public/dists/trixie/) and for [proxmox](https://packages.linbit.com/public/dists/proxmox-9/) which is also debian based and provides a newer kernel. Note that their website only provides [instructions for proxmox](https://linbit.com/drbd-user-guide/linstor-guide-1_0-en/#s-proxmox-installing-from-linbit-public-repos), but debian versions are also there.

I'm not sure what code it's compiled from, but it's probably a better base than the PPA for Ubuntu.

stgraber:
Ah, I somehow missed that repository. That's good, it should make things quite a bit easier to include then.

victoitor:
I just spent the last 2 hours looking everywhere for a reference to this repository and I can't find it anywhere. Wanted to see if I could find any comment on using release candidate code on it. I actually found it by opening the proxmox public repository link and browsing around.

PR #524, opened 2025-11-09
Add DRBD9 kernel module

stgraber:
Preparation for #436


mkosi.images/base/mkosi.conf.d/01-kernel-drbd.sh
```diff
@@ -0,0 +1,51 @@
...
+# Install the drbd sources
+mkdir -p /run/lock
+apt-get install --no-install-recommends --yes drbd-dkms || true
+
+# Patch the module source
+cp "/buildroot/boot/config-${KERNEL}" "/buildroot/usr/src/linux-headers-${KERNEL}/.config"
+cd /buildroot/usr/src/drbd-*/src/ || exit 1
+(
+cat << EOF
+ZGlmZiAtLWdpdCBhL2RyYmQvS2J1aWxkLmRyYmQgYi9kcmJkL0tidWlsZC5kcmJkCmluZGV4IDU5
```
gibmat, line 13: Any way to include this patch without embedding it as a base64-encoded chunk?

stgraber: Not that I could find. It was getting pretty ugly having to escape a bunch of stuff in there that was making the shell unhappy.

And I couldn't find a good way for mkosi to make the patch available only to the build environment.

stgraber: It's hopefully a temporary hack until the DRBD folks push out a release that works with 6.17.

Until then the patch forcefully turns off dax support.

gibmat (approved):

PR #530, opened 2025-11-10
incus-osd/services: Add the ability to expose Incus via Tailscale Serve

### LINKED ISSUE #529: Support exposing Incus (port 8443) via Tailscale Serve
Tailscale Serve has some nice HTTPS certificate management magic, and this would allow users to take advantage of that, and access their Incus server via `https://{hostname}.{tailnet}.ts.net/` with a valid HTTPS cert trusted by default on most clients. Saves having to deploy your own HTTPS Root CA.

quagsirus:
Closes #529

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -24,6 +24,7 @@ Furo
 FuturFusion
 GiB
 Github
+HTTPS
 Incus
 IncusOS
 IPs


incus-osd/internal/services/service_tailscale.go
```diff
@@ -134,5 +134,17 @@ func (n *Tailscale) configure(ctx context.Context, needsRejoin bool) error {
 		return err
 	}
 
+	_, err = subprocess.RunCommandContext(ctx, "tailscale", "serve", "reset")
+	if err != nil {
+		return err
+	}
+
+	if n.state.Services.Tailscale.Config.ServeEnabled {
+		_, err = subprocess.RunCommandContext(ctx, "tailscale", "serve", "--bg", "--https="+strconv.Itoa(int(n.state.Services.Tailscale.Config.ServePort)), "https+insecure://localhost:8443")
```
breml, line 148: How would the user do the authentication dance? If I remember correctly, tailscale prints an URL, the user needs to visit in order for the instance to be authenticated with the users tailnet. This url would be needed to be shown to the user in the IncusOS console.
Alternatively, an auth key could be used (https://tailscale.com/kb/1085/auth-keys), which is provided though the IncusOS seed configuration.

Edit: Ah I guess, the existing config from `incus-osd/api/service_tailscale.go` is re-used.

quagsirus: Yeah auth to the tailnet has already been handled, but your question has made me want to look into if `tailscale serve` requires any interactivity in different scenarios - it doesn't at first glimpse but I'm not fully certain

quagsirus:
Fixed the sign-off issue and Tailscale build error

incus-osd/internal/services/service_tailscale.go
quagsirus: This could happen if HTTPS certificates have not been enabled, and adding `--yes` unfortunately doesn't make it error out
image

quagsirus: There's nothing special about the link provided by the command though, and provided that "Enable HTTPS" at https://login.tailscale.com/admin/dns is clicked beforehand, there is no interactivity required
image

stgraber: Given the potential for an interactive prompt, a context with a short timeout should be used here. Probably something like a 5s timeout would do.

quagsirus: Agreed, 5s timeout added

quagsirus:
I've added to the gitignore, if you'd rather I didn't then I'm happy to drop/revert that commit

PR #534, opened 2025-11-12
Small typo

dwlfrth:


breml (approved):
LGTM, looks like the rest of the JSON would need some love as well to make it pretty formatted.

PR #539, opened 2025-11-12
doc/contributing: Add a debugging section

gibmat:


doc/contributing.md
```diff
@@ -32,3 +32,22 @@ To test the update process, build a new image and update to it with:
...
+
+    cd ./incus-osd/
+    go build ./cmd/incus-osd/
+    incus file push ./incus-osd test-incus-os/root/
+
+Then `exec` into the virtual machine, stop the main `incus-osd` service and run the local copy:
+
+    incus exec test-incus-os bash
+    systemctl stop incus-osd
+    /root/incus-osd
```
breml: The trick, which @stgraber showed me, also allows to still start the side-loaded version of the daemon with systemd:

```
    incus exec test-incus-os bash
    systemctl stop incus-osd
    mount -o bind /root/incus-osd /usr/local/bin/incus-osd
    systemctl start incus-osd
```

doc/contributing.md
breml, line 1: Since there is no editor inside of IncusOS, and therefore the `EDITOR` env var is not set, I found the following one-liner useful (to be executed inside of IncusOS):

```
mkdir /tmp/micro && curl -sLo /tmp/micro/micro.tar.gz https://github.com/zyedidia/micro/releases/download/v2.0.14/micro-2.0.14-linux64-static.tar.gz && tar -xzf /tmp/micro/micro.tar.gz -C /tmp/micro --strip-components=1 && mv /tmp/micro/micro /root/micro && rm -rf /tmp/micro && export EDITOR=/root/micro
```

doc/contributing.md
stgraber: Yeah, I generally prefer the side-loading approach as that works for any of the binaries including incusd, operations-centerd, ... and allows for the system to run otherwise normally.

doc/contributing.md
stgraber: Ah yeah, that could be useful to add

breml: Website: https://github.com/zyedidia/micro

Most useful shortcuts: 😉 

```
Ctrl+q to quit
Ctrl+s to save
```

PR #544, opened 2025-11-13
Allow setting system timezone

### LINKED ISSUE #477: Allow setting timezone for IncusOS
At the moment we default to UTC for IncusOS, but it might be useful if we had a way to allow the user to set their preferred timezone. This could help make defining update maintenance windows more intuitive.

gibmat:
* Rename the `SystemNetworkNTP` struct to `SystemNetworkTime`
* Add a `Timezone` field to `SystemNetworkTime`

Closes #477

### DIFF
--- a/doc/reference/system/network.md
+++ b/doc/reference/system/network.md
@@ -30,9 +30,9 @@ The following configuration options can be set:
 
 * `dns`: Optionally, configure custom DNS information for the system.
 
-* `ntp`: Optionally, configure custom NTP server(s) for the system.
-
 * `proxy`: Optionally, configure a proxy for the system.
+
+* `time`: Optionally, configure custom NTP server(s) and timezone for the system.
 
 ### Examples
 
@@ -93,7 +93,7 @@ Configure a VLAN with ID 123 on top of an active-backup bond composed of two int
 }
 ```
 
-Configure custom DNS and NTP for IncusOS:
+Configure custom DNS, NTP, and timezone for IncusOS:
 
 ```
 {
@@ -103,8 +103,9 @@ Configure custom DNS and NTP for IncusOS:
         "search_domains": ["example.com", "example.org"],
         "nameservers": ["ns1.example.com", "ns2.example.com"]
     },
-    "ntp": {
-        "timeservers": ["ntp.example.com"]
+    "time": {
+        "ntp_servers": ["ntp.example.com"],
+        "timezone": "America/New_York"
     }
 }
 ```


doc/reference/seed.md
```diff
@@ -86,3 +86,7 @@ This file provides preseed information to configure a given provider, which is u
 to fetch IncusOS updates and applications.
 
 The structure used is the [provider API struct](https://github.com/lxc/incus-os/blob/main/incus-osd/api/system_provider.go).
+
+### `timezone.{json,yml,yaml}`
+This file provides information to configure the system's timezone on first boot.
+The structure is defined in [`api/seed/timezone.go`](https://github.com/lxc/incus-os/blob/main/incus-osd/api/seed/timezone.go).
```
gibmat: The link check is failing because `api/seed/timezone.go` is part of this PR and doesn't yet exist in main.

stgraber:
@gibmat I don't love having a dedicated seed just for the timezone or having the action endpoint to set it without a good way to read it back.

I know it'd be slightly ugly, but what about having a TimeZone field in SystemNetworkNTP instead?

That would line up with the Hostname field from the SystemNetworkDNS which is similarly not strictly a network/DNS thing.

gibmat:
> @gibmat I don't love having a dedicated seed just for the timezone or having the action endpoint to set it without a good way to read it back.

Yeah, but I couldn't think of any other way to have it so a freshly-installed system comes up and sets its timezone properly. Requiring a post-install API call seems sub-optimal, especially if someone is trying to deploy a lot of IncusOS servers.

I did add the system's timezone to the `/1.0` API response as part of the map of environment values.

stgraber:
> > @gibmat I don't love having a dedicated seed just for the timezone or having the action endpoint to set it without a good way to read it back.
> 
> Yeah, but I couldn't think of any other way to have it so a freshly-installed system comes up and sets its timezone properly. Requiring a post-install API call seems sub-optimal, especially if someone is trying to deploy a lot of IncusOS servers.
> 
> I did add the system's timezone to the `/1.0` API response as part of the map of environment values.

If we make it a `TimeZone` option within the network config, then we'd have a way to seed it and to edit it over the API.

incus-osd/api/system_network.go
```diff
@@ -34,6 +34,8 @@ type SystemNetworkConfig struct {
 	Interfaces []SystemNetworkInterface `json:"interfaces,omitempty" yaml:"interfaces,omitempty"`
 	Bonds      []SystemNetworkBond      `json:"bonds,omitempty"      yaml:"bonds,omitempty"`
 	VLANs      []SystemNetworkVLAN      `json:"vlans,omitempty"      yaml:"vlans,omitempty"`
+
+	Timezone string `json:"timezone,omitempty" yaml:"timezone,omitempty"`
```
stgraber: Feels very weird having it there on its own. I was thinking of putting it inside of `NTP` as it already has `TimeServers` there.

Maybe just rename `NTP`and its struct to `Time`, that way TimeZone feels less odd in there and it gives us an option to add PTP settings in there eventually?

gibmat: Yeah, without renaming the struct it would feel quite hacky, especially if a user only wanted to set the timezone.

Are we OK with that breakage in the API now?

stgraber: Yeah, I'm fine with breaking that particular corner of the API, provided we have the correct state migration logic in place.

stgraber:
@gibmat `doc/reference/network` needs updating, otherwise this is good to go.

stgraber:
@gibmat you included your notes document in there ;)

gibmat:
> @gibmat you included your notes document in there ;)

Oops! :laughing:

PR #549, opened 2025-11-14
Add Linstor satellite support

### LINKED ISSUE #436: Add linstor support

### LINKED ISSUE #525: Use separate applications for Ceph and Linstor support
Both Ceph and Linstor are quite large as they pull their own languages (python3 for Ceph, java for Linstor).

Rather than inflict that on everyone, it'd probably make sense to have this layout:
- base
  - incus
    - incus-ceph
    - incus-linstor

So adding another layer of overlay for the tools of both of those distributed storage options.

@gibmat do you expect us to run into any issue if doing that? I'm think manifest/changelog type of issues.
We'll definitely need a few tweaks to IncusOS to let us:
 - Easily add an application
 - Be aware that `incus-ceph` and `incus-linstor` can only be installed on systems running `incus`
 - Tweak the service checks for `ceph` and `linstor` to check that the application is installed

Current compressed image size is at 288MB, adding Linstor would bump to 420MB and I vaguely remember Ceph already cause a bump of 80MB or so. Splitting things definitely feels worth it to limit the amount downloaded by users.

stgraber:
Closes #436
Closes #525

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,6 +96,8 @@ jobs:
 
           mv mkosi.output/debug.raw upload/
           mv mkosi.output/incus.raw upload/
+          mv mkosi.output/incus-ceph.raw upload/
+          mv mkosi.output/incus-linstor.raw upload/
           mv mkosi.output/migration-manager.raw upload/
           mv mkosi.output/operations-center.raw upload/
 


stgraber:
@gibmat similar to your PR, the doc failure is because of a link which can't yet be resolved

stgraber:
@gibmat this should be good to go

Makefile
```diff
@@ -164,6 +164,8 @@ test-applications:
 
 	incus file push mkosi.output/debug.raw test-incus-os/root/updates/
 	incus file push mkosi.output/incus.raw test-incus-os/root/updates/
+	incus file push mkosi.output/incus-ceph.raw test-incus-os/root/updates/
```
gibmat, line 167: Need to also add the two new application to `.github/workflows/build.yaml` to actually publish them. :)

incus-osd/internal/rest/api_applications.go
```diff
@@ -46,34 +47,104 @@ import (
...
+		err := json.NewDecoder(r.Body).Decode(app)
+		if err != nil {
+			_ = response.BadRequest(err).Render(w)
+
+			return
+		}
+
+		// Input validation.
+		if app.Name == "" {
+			_ = response.BadRequest(err).Render(w)
```
gibmat: Shouldn't use `err` here, but something like `errors.New("invalid application name '" + app.Name + "'")`

incus-osd/internal/rest/api_applications.go
```diff
@@ -46,34 +47,104 @@ import (
...
+
+			return
+		}
+
+		_, ok := s.state.Applications[app.Name]
```
gibmat: Naming the variable `ok` is a bit confusing

incus-osd/internal/rest/api_applications.go
```diff
@@ -46,34 +47,104 @@ import (
...
+//	  - application/json
+//	parameters:
+//	  - in: body
+//	    name: application
+//	    description: Application to be installed
+//	    required: true
+//	    schema:
+//	      type: object
+//	      example: {"name": "incus"}
+//	responses:
```
gibmat, line 71: Missing `Conflict` response in the list

mkosi.images/incus-ceph/mkosi.conf
```diff
@@ -0,0 +1,12 @@
+[Config]
+Dependencies=base
```
gibmat, line 2: Minor nit -- I think that both the ceph and linstor application images could be switched to depending on the incus layer, since that's how they will be used, and it may reduce duplication of common dependencies. (Not sure how much this would reduce image sizes, and I haven't tried it locally.)

stgraber: Nope, that doesn't work, I tried :)

Basically you need a full fledged image, not a sysext as the base, otherwise you don't get the working base directory which can be layered and the build fails.

gibmat: Darn :slightly_frowning_face:

Issue #559, opened 2025-11-15
failed to check for secure boot key updates


mybigman:
Trying to run off 64GB usb flash drive.

Loaded the keys in the secure boot bios from the usb flash drive.

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/b9e42516-80c6-43e3-a69a-bddaef3880a7" />

Using `https://incusos-customizer.linuxcontainers.org/ui/`

```
Image type: USB
Image architecture: x86_64
Image usage: Operation
Image application: Incus
Installation target: nothing checked
Application configuration: incus remote get-client-certificate
```

Using `https://store.minisforum.com/products/minisforum-motherboard`
```
MINISFORUM BD795i SE
```

stgraber:
Is your BIOS up to date? 

The error suggests something pretty weird going on with the UEFI NVRAM here. It could be a firmware issue of some kind.

gibmat:
> The error suggests something pretty weird going on with the UEFI NVRAM here. It could be a firmware issue of some kind.

Yeah, after verifying you've got the latest BIOS, I'd also be interested in seeing what Secure Boot key(s) are enrolled for the PK, KEK, and db EFI variables. You should be able to manually view these from the BIOS' Secure Boot configuration screen(s).

mybigman:
Bios is using the latest `1.12`

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/a21d95a3-954a-4402-a3ba-f1a4427bc711" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/4f7bdcfe-7a4c-4970-86e5-28b3284bbde5" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/3815cd01-361c-48db-a550-5f4cad329179" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/f46e91a2-6b5d-4158-ae5c-9bc76775bf96" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/afdd9875-3cc4-47ff-a718-a55ac556ca10" />

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/110f5ac9-33a6-4f18-9087-2d06ca9b54b5" />

stgraber:
Can you try kicking out all the KEKs except for the IncusOS one?

mybigman:
[...]

PR #567, opened 2025-11-17
Include underlying error when we cannot read an EFI variable

gibmat:
Will help debugging issues like #559.


incus-osd/internal/secureboot/efi_vars.go
```diff
@@ -131,7 +131,7 @@ func UpdateSecureBootCerts(ctx context.Context, tarArchive string) (bool, error)
 func applySecureBootUpdates(ctx context.Context, varName string, newCerts map[string][]byte) (bool, error) {
 	existingCerts, err := GetCertificatesFromVar(varName)
 	if err != nil {
-		return false, fmt.Errorf("failed to read EFI variable '%s'", varName)
+		return false, fmt.Errorf("failed to read EFI variable '%s' (%s)", varName, err.Error())
```
stgraber: Why the odd formatting?

Wouldn't it be better to do `fmt.Errorf("failed to read EFI variable %q: %w", varName, err)`?

stgraber (approved):

PR #570, opened 2025-11-17
Allow simple RAID0 and RAID1 pools for "local" default ZFS pool

### LINKED ISSUE #532: Add "local" ZFS pool RAID configuration options
Several users have said it would be useful if the default "local" ZFS pool could be configured with another local drive, either as RAID-0 or RAID-1. This is technically feasible, but will introduce some complexity:

- Some new install seed values will be needed. Minimally, a selection of RAID-0 or RAID-1 and the second drive to use when the pool is initially created.
- The zpool on the main system drive will be ~35GiB smaller than the disk, due to the other partitions. Need to determine how to best handle this when adding the second drive to the pool.
- It will only be possible to replace the second drive if it exhibits failures.
- Need to think about how recovery/reinstall and importing the second drive of an existing "local" ZFS pool.

The "local" ZFS pool won't be allowed to consist of more than two drives or use other more complex setups. Users needing that should create dedicated zpools independent of the "local" pool.

gibmat:
Closes #532

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -69,6 +69,7 @@ proxied
 Proxmox
 Proxmox's
 raidz
+resilver
 RSA
 SLAAC
 struct


stgraber:
I'm really not sure I'd want to do this through the seed. There is no strong need for us to provision it that early when it can be handled later on.

Instead I'd much rather we allow the expansion of the local pool to another device after the fact through the normal storage API. This will then also allow us the normal set of operations on it, including the replacement of the defective secondary drive.


So we'd effectively be allowing:
 - Addition of a second device to the local pool
 - As part of the addition of that second device, allow changing the type to `zfs-raid0` or `zfs-raid1` (the others can't be done going from a standalone drive)
 - The second device would be provided as a whole disk, if `zfs-raid0`, it will be added as a whole to the local pool (extended). If `zfs-raid1`, we will partition the drive and put the ZFS partition at the same offset as the root disk, then set up the mirror.

incus-osd/internal/zfs/zfs.go
```diff
@@ -419,6 +430,94 @@ func UpdateZpool(ctx context.Context, newConfig api.SystemStoragePool) error {
...
+	}
+
+	if len(newConfig.Log) > 0 {
+		return errors.New("special zpool 'local' cannot have any log devices")
+	}
+
+	if len(newConfig.Cache) > 0 {
+		return errors.New("special zpool 'local' cannot have any cache devices")
+	}
+
```
stgraber, line 478: Any specific reason why not?

gibmat: Mostly a desire to keep the "local" pool as simple as possible. :slightly_smiling_face:

PR #572, opened 2025-11-18
Rework zpool import logic

### LINKED ISSUE #571: Rework ZFS pool import logic
We currently run `zpool import -a` so we can list all pools, even those we don't manage, we then have a `:import-encryption-key` route to load a missing key and get a pool up and running.

This causes some confusion for users who have additional pools which are not meant for IncusOS to see, such as disks that will be passed in their entirety to a VM.

To sort this out, I think we should:
 - Only do specific imports (`zpool import NAME`)
 - Replace the current `:import-encryption-key` API route with a `:import-pool` route instead which would allow for providing the pool name and its encryption key, getting it imported and added to config.

gibmat:
Closes #571


stgraber:
@gibmat needs a quick rebase but otherwise good to go

gibmat:
> @gibmat needs a quick rebase but otherwise good to go

Made a few adjustments to handle changes from #570 and added another test; should be good to go now.

PR #574, opened 2025-11-19
Improve the debug application

stgraber:


doc/contributing.md
```diff
@@ -53,11 +53,7 @@ Then `exec` into the virtual machine, stop the main `incus-osd` service and run
...
-
-    mkdir /tmp/micro/
-    curl -sLo /tmp/micro/micro.tar.gz https://github.com/zyedidia/micro/releases/download/v2.0.14/micro-2.0.14-linux64-static.tar.gz
-    tar -xzf /tmp/micro/micro.tar.gz -C /tmp/micro --strip-components=1
-    mv /tmp/micro/micro /root/micro
-    rm -rf /tmp/micro
-    export EDITOR=/root/micro
+When debugging, it's a good idea to install the `debug` application which contains a variety of useful tools, including a basic text editor (`nano`).
+
+    incus exec test-incus-os bash
```
gibmat, line 58: Missing a `make test-applications` to push the application images into the VM. I'd also recommend an explicit `http://localhost` rather than `os` in the curl command to reduce potential confusion:

```
make test-applications
incus exec test-incus-os bash
curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/applications -X POST -d '{"name": "debug"}'
```

stgraber: That section isn't really written for the case where you build your own image as in that situation you wouldn't need to side-load `incus-osd` in the first place.

stgraber: For `curl` I lined it up with how we document it in the Incus documentation.

I actually find using `http://localhost/XYZ` to be worse as that suggests that we are going over the network (`localhost`) when we're not. Purposefully omitting that part of the URL makes it clearer that something odd is going on (going over a unix socket).

PR #585, opened 2025-11-20
Implement storage volume management

stgraber:


stgraber:
@gibmat the upgrade logic will obviously fail in tests and won't really work in production either as this would run before the pool is imported.

I'm thinking of having the state expose both the current schema version and the previous one so that when dealing with an upgrade we can later on in the code what version we've just upgraded to and apply the needed tweak. What do you think?

incus-osd/api/system_storage.go
```diff
@@ -31,14 +31,23 @@ type SystemStoragePool struct {
...
+	RawPoolSizeInBytes        int                       `json:"raw_pool_size_in_bytes"        yaml:"raw_pool_size_in_bytes"`
+	UsablePoolSizeInBytes     int                       `json:"usable_pool_size_in_bytes"     yaml:"usable_pool_size_in_bytes"`
+	PoolAllocatedSpaceInBytes int                       `json:"pool_allocated_space_in_bytes" yaml:"pool_allocated_space_in_bytes"`
+	Volumes                   []SystemStoragePoolVolume `json:"volumes"                       yaml:"volumes"`
+}
+
+// SystemStoragePoolVolume represents a single IncusOS-managed volume in a pool.
+type SystemStoragePoolVolume struct {
+	Name         string `json:"name"           yaml:"name"`
+	UsageInBytes int64  `json:"usage_in_bytes" yaml:"usage_in_bytes"`
```
gibmat: We've been using `int` elsewhere in the API; is there a specific reason for using `int64`?

incus-osd/internal/rest/api_system_storage.go
```diff
@@ -331,3 +333,103 @@ func (*Server) apiSystemStorageImportPool(w http.ResponseWriter, r *http.Request
...
+	if r.Method != http.MethodPost {
+		_ = response.NotImplemented(nil).Render(w)
+
+		return
+	}
+
+	type createStruct struct {
+		Pool string `json:"pool"`
+		Name string `json:"name"`
+		Size int64  `json:"size"`
```
gibmat: Call this `Quota` to be consistent?

doc/reference/system/storage.md
```diff
@@ -90,3 +90,26 @@ If importing an existing storage pool, IncusOS needs to be informed of its encry
...
 ```
+
+## Managing volumes
+
+It's possible to create and delete volumes within a storage pool.
+
+Each volume has its own:
+
+- Name
+- Quota (in bytes)
```
gibmat: Add a note about `0` meaning no quota

incus-osd/api/system_storage.go
stgraber: Nah, I can change it to int, Incus tends to use int64 because we actually cared about 32bit architectures and int32 is a bit small when dealing with large number of bytes.

gibmat:
> @gibmat the upgrade logic will obviously fail in tests and won't really work in production either as this would run before the pool is imported.
> 
> I'm thinking of having the state expose both the current schema version and the previous one so that when dealing with an upgrade we can later on in the code what version we've just upgraded to and apply the needed tweak. What do you think?

It feels like that would be overloading the state's version field. Could we add a `System.Storage.Upgrade` field or similar to the state struct that we could consult once the pools are imported?

Alternatively, maybe we could just always apply the update to the `incus/local` pool at boot for a month or two. Setting zfs properties are idempotent, so re-applying the `incusos:use` role wouldn't error out. Once the existing IncusOS installs have had a chance to upgrade their storage, we can eventually remove that bit of upgrade logic. The advantage of this would be that we wouldn't have to maintain long-term upgrade logic for storage.

incus-osd/api/system_storage.go
gibmat: Yeah, no problem with that for IncusOS :grinning:

stgraber:
> > @gibmat the upgrade logic will obviously fail in tests and won't really work in production either as this would run before the pool is imported.
> > I'm thinking of having the state expose both the current schema version and the previous one so that when dealing with an upgrade we can later on in the code what version we've just upgraded to and apply the needed tweak. What do you think?
> 
> It feels like that would be overloading the state's version field. Could we add a `System.Storage.Upgrade` field or similar to the state struct that we could consult once the pools are imported?
> 
> Alternatively, maybe we could just always apply the update to the `incus/local` pool at boot for a month or two. Setting zfs properties are idempotent, so re-applying the `incusos:use` role wouldn't error out. Once the existing IncusOS installs have had a chance to upgrade their storage, we can eventually remove that bit of upgrade logic. The advantage of this would be that we wouldn't have to maintain long-term upgrade logic for storage.

Yeah, I'll probably go with that approach, leaving a comment in there to remove that code in early 2026.

PR #596, opened 2025-11-24
app-build: Update dependencies

stgraber:
Most importantly, this now pins Migration Manager to v0.1.0.

### DIFF
--- a/app-build/applications.json
+++ b/app-build/applications.json
@@ -37,7 +37,7 @@
         ]
     },
     "migration-manager": {
-        "version": "main",
+        "version": "0.1.0",
         "repo": "https://github.com/FuturFusion/migration-manager.git",
         "build_targets": [
             [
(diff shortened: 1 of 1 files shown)


gibmat:
I was also thinking of bumping the build of IncusOS to use go 1.25?

stgraber:
@gibmat you mean the go.mod minimum version?

I don't really see a reason for bumping it. We'll always build with the latest stable Go anyway. The only affect of bumping the Go version in go.mod is to force all our downstreams to also bump theirs, which I really try to avoid until it's necessary.

stgraber:
Incus has a policy of always supporting the last two stable releases, so currently 1.24 and 1.25. If we bump IncusOS to 1.25, we'll need to have Incus pin an old tag of IncusOS until Go 1.26 is released.

So at least for the inter-connected projects (Incus, IncusOS, Migration Manager, Operations Center), we likely want to stick to the same policy of the last two minor Go releases.

gibmat:
Sure, that makes sense.

gibmat (approved):

PR #598, opened 2025-11-24
Add tutorial for preparing Incus storage pool

### LINKED ISSUE #584: Document ZFS pool creation through IncusOS and adding them to Incus
Hi 
After last update I receive these warnings. System seems to be working..... and storage are listed.
Should I be worried about this? Thank you.

Image

Image

gibmat:
Closes #584


stgraber:
@gibmat We should recommend using the new storage volume feature, so:
 - Create the pool in IncusOS
 - Create a volume on it for use by Incus
 - Create the storage pool in Incus using it

That will clean up the messy `state` you see at the end of your example and give us a chance to properly handle config tweaks and optimizations in the future.

PR #602, opened 2025-11-24
Fix interaction of migration windows and configured update check frequency

### LINKED ISSUE #531: Better handle interaction of update windows and check interval
Currently it's possible to define an update window (say, for example daily from 2am - 4am) and the default check interval of six hours will result in IncusOS never actually being able to check for an update _except_ if you happen to have initially booted the system sometime between 2am - 4am.

We should process maintenance windows before sleeping, and then sleep the minimum time of the configured check interval and the next maintenance window (if any).

gibmat:
Closes #531

### DIFF
--- a/incus-osd/cmd/incus-osd/main.go
+++ b/incus-osd/cmd/incus-osd/main.go
@@ -573,8 +573,19 @@ func updateChecker(ctx context.Context, s *state.State, t *tui.TUI, p providers.
 				break
 			}
 
+			// If any maintenance windows are defined, limit the time to sleep to be a minimum
+			// of the configured check frequency and the start of the next maintenance window,
+			// whichever is shorter.
+			for _, window := range s.System.Update.Config.MaintenanceWindows {
+				if window.TimeUntilActive() > 0 && window.TimeUntilActive() < frequency {
+					frequency = window.TimeUntilActive()
+				}
+			}
+
 			if timeSinceCheck < frequency {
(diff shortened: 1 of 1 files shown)


incus-osd/api/system_update.go
```diff
@@ -43,8 +43,58 @@ func (w *SystemUpdateMaintenanceWindow) IsCurrentlyActive() bool {
...
+	if endOffset >= 7*24*60 && int(t.Weekday()) <= int(w.EndDayOfWeek.ToWeekday()) {
+		currentPosition += 7 * 24 * 60
+	}
+
+	// Migration window is currently active.
+	if startOffset <= currentPosition && currentPosition <= endOffset {
+		return 0 * time.Minute
+	}
+
+	// Migration window hasn't yet started.'
```
stgraber: You've got a random `'` here :)

stgraber (approved):

incus-osd/api/system_update.go
gibmat: Yeah, I wanted to push my WIP yesterday afternoon before taking a bit of a mental break. :slightly_smiling_face:

stgraber:
@gibmat is there more needed or is that good to go?

gibmat:
> @gibmat is there more needed or is that good to go?

Still a little bit of tweaking

gibmat:
@stgraber OK, I think this should be ready now.

PR #617, opened 2025-11-26
Add certificate generation to the customizer

stgraber:


incus-osd/cmd/image-customizer/html/index.html
```diff
@@ -179,15 +179,15 @@ <h2 class="accordion-header">
                             </div>
 
                             <div class="mb-3">
-                                <label class="form-label" for="applicationClientCertificate">TLS certificate of the initial trusted client.<br>This can be retrieved by running <code>incus remote get-client-certificate</code>.</label>
+                                <label class="form-label" for="applicationClientCertificate">
+                                    <b>TLS client certificate of the initial trusted client.</b>
+                                    <br>
+                                    If using the <a href="https://github.com/lxc/incus/releases/latest/" target="_blank">Incus CLI tool</a>, this can be retrieved by running <code>incus remote get-client-certificate</code>.
+                                    <br>
+                                    Alternatively, one can be <a onclick="certificate()" href="#certificate">generated</a> for you which can be imported in your web browser for UI access.
```
gibmat, line 187: Should we have a warning or something about it being more secure to use a certificate you generate, rather than relying on one that the image customizer produces? Since that way there's no way for the private key to be exposed, even though we don't save any of the generated credentials?

stgraber: I've added a note to the modal dialog to point out that a locally generated certificate is safer.

stgraber:
I added the cert generation logic directly to the codebase now so we have a bit of control on the CN as that's displayed pretty prominently in web browsers.

PR #642, opened 2025-11-30
feat(image-customizer): select release channel

theoludwig:
Add a way to select between the `testing` and the `stable` release channel, while keeping the `stable` channel selected by default for the IncusOS image downloader: https://incusos-customizer.linuxcontainers.org/ui/

image

(tested with `go run main.go "~/Downloads/incus"`)

### DIFF
--- a/incus-osd/cmd/image-customizer/html/index.html
+++ b/incus-osd/cmd/image-customizer/html/index.html
@@ -202,30 +202,52 @@
 
                     <div id="panelAdvanced" class="accordion-collapse collapse">
                         <div class="accordion-body">
-                            <h3>Network configuration</h3>
+                            <div class="mb-3">
+                                <h3>Network configuration</h3>
 
-                            <p>
-                                By default IncusOS will run both DHCPv4
-                                and SLAAC on every interface it can find
-                                on the system, hoping that one of them
-                                will provide enough connectivity to
-                                allow the user to connect and apply the
-                                final network configuration.
-                            </p>
+                                <p>
+                                    By default IncusOS will run both DHCPv4
+                                    and SLAAC on every interface it can find
+                                    on the system, hoping that one of them
+                                    will provide enough connectivity to
+                                    allow the user to connect and apply the
+                                    final network configuration.
+                                </p>
 
-                            <p>
-                                For cases where this approach doesn't
-                                work, such as environments requiring
-                                static IP addresses, the use of a
-                                proxy server or using link aggregation
-                                or VLANs, a full network configuration
-                                can be provided instead.
-                            </p>
+                                <p>
+                                    For cases where this approach doesn't
+                                    work, such as environments requiring
+                                    static IP addresses, the use of a
+                                    proxy server or using link aggregation
+                                    or VLANs, a full network configuration
+                                    can be provided instead.
+                                </p>
 
-                            <p>Supported options and examples can be <a href="https://linuxcontainers.org/incus-os/docs/main/reference/system/network/" target="_blank">found here</a>.</p>
+                                <p>
+                                    Supported options and examples can be <a href="https://linuxcontainers.org/incus-os/docs/main/reference/system/network/" target="_blank">found here</a>.
+                                </p>
+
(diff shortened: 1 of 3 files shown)


stgraber:
Hmm, so I'm not actually so sure of the usefulness of this one.

That's because downloading an image from the "testing" channel doesn't actually lead to a system running from the "testing" channel.

For that you still need to use `incus admin os system update edit` and change the channel of the system and then check for any update.

The only case where downloading from testing is appropriate is when debugging installation or early boot issues, basically anything before you'd be able to run `incus admin os system update edit`.

This is pretty rare and is something that's generally just done by the developers or by our testing infrastructure and in both of those cases, we're unlikely to use the download website for that.

theoludwig:
Thank you for the fast reply! @stgraber 
I just discovered the Incus project, and started experimenting with it, so far, I find it amazingly good, I might even migrate a Proxmox server I have to IncusOS instead. :smile:

> That's because downloading an image from the "testing" channel doesn't actually lead to a system running from the "testing" channel.

That sounds a fine behavior to me.
From my understanding of your release process (please, correct me, if I'm wrong): you regularly release new "testing" version, and when you think it's the good moment, you take a already released "testing" version, and you promote it to stable, is that correct?

If yes, my expectation is that by installing the "testing" iso, I will for example use the `202511292320` version, and then when later, for example, the `202512182320` version is available to stable, I can update to it.

But I understand the confusion that it can creates, either we explain it better, or we want to avoid this "hassle", and don't want to add this feature to the  IncusOS image downloader site.

---

Story time to explain why I wanted to use the "testing" ISO. :laughing: 

So, I wanted to start Docker containers inside LXC containers, but was impacted by this bug: 
[CVE-2025-52881: fd reopening causes issues with AppArmor profiles (`open sysctl net.ipv4.ip_unprivileged_port_start file: reopen fd 8: permission denied`)](https://github.com/opencontainers/runc/issues/4968) 

which was apparently fixed in https://github.com/lxc/incus/pull/2624 and is included in v6.19 of Incus, so instead of using Incus 6.18 which was what included in the latest stable ISO, I wanted to use Incus 6.19.

(somehow using Incus 6.19.1 was not enough, I also needed to do `apt install --yes --allow-downgrades containerd.io=1.7.28-1~debian.13~trixie` inside the Debian 13 container to make it work)

stgraber:
Ah, interesting and good that this worked for you. For extra context on how things work and extra fun that can happen when installing from a testing image without updating the channel in the update config.

IncusOS doesn't actually come with Incus installed. The application you select at download time instead gets installed on first boot. 

We used to respect the channel config so you would have gotten the stable version of Incus even though the base OS would be testing.

Luckily for you, we did tweak that logic since so that if the OS version is higher than anything available in the channel on the image server, it will go and pick an application of the same version as the OS.

Now for the problem. We only keep the most recent 3 testing images. So if your system isn't fully installed within the time we tag the next 3 testing images, it will completely fail to boot.

stgraber:
Anyway, we can add a checkbox in the advanced section for it. 

Something like:
[ ] Use an experiential (testing channel) image

Note that your current commit is missing the required Signed-off-by line before we can run tests and review its content in detail.

theoludwig:
> Anyway, we can add a checkbox in the advanced section for it.
> 
> Something like: [ ] Use an experiential (testing channel) image
> 
> Note that your current commit is missing the required Signed-off-by line before we can run tests and review its content in detail.

Alright, the commit message should now include "Signed-off-by" (I force-pushed). :+1:
And I moved this section to the "Advanced settings" accordion.

image

I kept using radio buttons instead of a checkbox, because that way I don't need to change the logic in the go-side + at least it makes it explicit, that you are using either "stable" or "testing". "Stable" is still selected by default of course.

Hope it's fine. :smile:

stgraber:
Thanks, I'll take it from there!

stgraber:
Updated the branch with:
 - Some re-formatting of the advanced section
 - Made the channel field a text field (as the channel names are server dependents and we have other environments with the customizer that use different names)
 - Simplified the server side logic to just go with what's being asked (can lead to a download error but that's fine)

stgraber (approved):

PR #646, opened 2025-12-01
incus-osd/systemd: Don't fail on misisng link speed

stgraber:


gibmat (approved):

Menthalion:
Could this have had a regression ? During an initial install of IncusOS I see the right adapter get the right IP adress, and then after an error message on the /speed file lose it again, and restart the network configuration over and over again.

Installing Ubuntu 26.04 or Debian 13.5 with the same network configurations (basically fixed ipv4 adress with /22 CIDR matched on mac address + a gateway) work, and show the correct speed in /sys/class/net/ens3f1/speed . IncusOS throws the error on `READ /sys/class/net/_hex_uuid/speed : invalid argument` and then restarts network config, looping into the same error.

Adapter is a BroadCom NetXtreme II BCM57810 10 Gigabit Ethernet

Configuration:
```
interfaces:
- name: "ens3f1"
  hwaddr: "redacted"
  
  addresses:
  - "10.0.32.10/22"
  
  routes:
  - to: "0.0.0.0/0"
    via: "10.0.35.254"

dns:
  nameservers:
  - "1.1.1.1"
  - "8.8.8.8"  

time:
  timezone: "Europe/Amsterdam"
```

stgraber:
Just rechecked and we only have the `speed` file logic in one spot and still handles the file being missing. In your case however the file isn't missing, it's instead broken (`invalid argument`) which is why it's still failing.

stgraber:
Please file a separate issue for that, but it's a bit of a trickier one for us to resolve as `invalid argument` may be a legitimate error that we should treat as such, masking it doesn't seem ideal even if that may be the only path forward here...

PR #655, opened 2025-12-02
Allow running in restricted MAC environments

stgraber:


incus-osd/api/system_network.go
```diff
@@ -47,6 +47,7 @@ type SystemNetworkInterface struct {
 	Hwaddr            string               `json:"hwaddr"                        yaml:"hwaddr"`
 	Roles             []string             `json:"roles,omitempty"               yaml:"roles,omitempty"`
 	LLDP              bool                 `json:"lldp"                          yaml:"lldp"`
+	StrictHwaddr      bool                 `json:"strict_hwaddr"                 yaml:"strict_hwaddr"`
```
gibmat: Would this be needed on a bond too?

incus-osd/internal/rest/api_system_network.go
```diff
@@ -124,6 +125,14 @@ func (s *Server) apiSystemNetwork(w http.ResponseWriter, r *http.Request) {
 
 		slog.InfoContext(r.Context(), "Applying new network configuration")
 
+		err = nftables.ApplyHwaddrFilters(r.Context(), newConfig.Config)
```
gibmat, line 128: This should be after `ApplyNetworkConfiguration`, otherwise there's a chance the interfaces might not be properly (re)named before trying to do stuff with `nft`.

Alternatively, just add this logic to the end of that method, to ensure it's always run properly when the network is configured?

stgraber: No, the interface doesn't need to exist for nft and we don't want to have any time where the interface exists and this config isn't applied or we'll get the network port shutdown at the switch level.

incus-osd/api/system_network.go
stgraber: I don't think so. On a bond, you're in a situation where the MAC will change anyways, so the switch is unlikely to have a port shutdown policy based on the MAC.

stgraber:
Resolved the merge conflicts

gibmat (approved):

benjamin-kirkbride:
For posterity, this is what you need if you get a scary email from Hetzner: `strict_hwaddr: true`

This could be better documented.

PR #660, opened 2025-12-03
Handle invalid Secure Boot certificates

gibmat:
Make IncusOS more resilient to bad OEM Secure Boot certificates:
  * Allow negative serial numbers
  * Warn about RSA keys larger than 2048 bits
  * Warn about non-RSA keys

Ideally these invalid certificates could be purged from the system, but sometimes they are required for hardware to work.

image

### DIFF
--- a/incus-osd/cmd/incus-osd/main.go
+++ b/incus-osd/cmd/incus-osd/main.go
@@ -1,3 +1,5 @@
+//go:debug x509negativeserial=1
+
 // Package main is used for the incus-osd daemon.
 package main
 
--- a/incus-osd/internal/secureboot/efi_vars.go
+++ b/incus-osd/internal/secureboot/efi_vars.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"context"
+	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/hex"
@@ -36,6 +37,24 @@ func GetCertificatesFromVar(varName string) ([]x509.Certificate, error) {
 	}
 
 	certs, _, err := parsedVal.SignatureData()
+
+	// Some consumer-grade manufacturers ship invalid Secure Boot certificates. Unfortunately, sometimes
+	// they are required for hardware such as storage or video cards to work. Since they are already in
+	// the PK/KEK/db EFI variable we still process them, but log a warning each time we do.
+	for _, cert := range certs {
+		if cert.SerialNumber.Sign() < 0 {
+			slog.Warn(fmt.Sprintf("Secure Boot variable '%s' contains invalid certificate '%s': serial number is negative", varName, cert.Subject))
+		}
+
+		switch publicKey := cert.PublicKey.(type) {
+		case *rsa.PublicKey:
+			if publicKey.Size()*8 > 2048 {
+				slog.Warn(fmt.Sprintf("Secure Boot variable '%s' contains invalid certificate '%s': RSA key length %d is greater than 2048 bits", varName, cert.Subject, publicKey.Size()*8))
+			}
+		default:
+			slog.Warn(fmt.Sprintf("Secure Boot variable '%s' contains invalid certificate '%s': expected a RSA key, got %s", varName, cert.Subject, cert.PublicKeyAlgorithm))
+		}
+	}
 
 	return certs, err
 }


cyyynthia:
Hello, sorry if this isn't the right place to raise this up but I'm neck deep in UEFI shenanigans and stumbled across this PR, so... >:D

> bad OEM Secure Boot certificates (...) RSA keys larger than 2048 bits

I'm having a hard time finding any definitive, authoritative information about the cryptographic requirements of UEFI. As RSA 2048 is to be banned in some countries (or at least deprecated e.g. in the U.S.) by 2030, I was trying to find information about support of stronger cryptography.

UEFI 2.10 *did* allow larger RSA keys up to 4096, and even allowed elliptic curve cryptography using NIST curves. https://uefi.org/specs/UEFI/2.10_A/32_Secure_Boot_and_Driver_Signing.html#firmware-os-crypto-algorithm-exchange

Somehow, this was deprecated and removed in UEFI 2.11. As I understand the spec, this leaves a blurry hole where there is no mention of cryptographic primitives that are allowed, only that certificates "should be at least RSA 2048". Which is an outdated requirement in the present day... UEFI PI 1.9 mentions the stronger algorithms UEFI 2.10 introduced, but in the context of signed firmware volumes only (i think?) ... :/

gibmat:
> I'm neck deep in UEFI shenanigans and stumbled across this PR, so... >:D

You have my condolences -- digging through the details of UEFI specs vs actual implementations is, uh, fun. :upside_down_face: 

> UEFI 2.10 _did_ allow larger RSA keys up to 4096, and even allowed elliptic curve cryptography using NIST curves. https://uefi.org/specs/UEFI/2.10_A/32_Secure_Boot_and_Driver_Signing.html#firmware-os-crypto-algorithm-exchange

Huh, that's interesting! I've only been looking at the current version of the UEFI spec (2.11); in this case specifically section 32.4, "Firmware/OS Key Exchange: Passing Public Keys". I wonder why crypto agility was depreciated in the current version?

> Somehow, this was deprecated and removed in UEFI 2.11. As I understand the spec, this leaves a blurry hole where there is no mention of cryptographic primitives that are allowed, only that certificates "should be at least RSA 2048". Which is an outdated requirement in the present day... UEFI PI 1.9 mentions the stronger algorithms UEFI 2.10 introduced, but in the context of signed firmware volumes only (i think?) ... :/

I agree that this does introduce some (additional) fuzziness when trying to interpret the spec. I did some quick poking but couldn't find any way to extract the actual UEFI spec version that an implementation is using. If we could get that, I could improve the warning message if someone is running IncusOS on a system that implements UEFI 2.10.

EDK II is the reference implementation for UEFI, and is used by Incus for UEFI VMs, so that's primarily what IncusOS is developed against. It does accept RSA-4096 certificates, so maybe that's actually allowed but not recommended because of the deprecated status?

Another related "de-facto" standard is that the Secure Boot update Microsoft's been rolling out over the past several months is a 2048 bit RSA cert. It's not uncommon for hardware vendors to only test against Microsoft OSes, and we've encountered issues with that in the past, so if certs larger than 2048 bits were used, I wouldn't be surprised to encounter issues with UEFI implementations unable to handle them.

cyyynthia:
> You have my condolences

Thank you 🤣

> It does accept RSA-4096 certificates, so maybe that's actually allowed but not recommended because of the deprecated status?

I initially tried to look at the problem looking for a list of supported algorithms. Now, I'm looking at it differently and asking whether there is anything that would prohibit use of anything that isn't RSA 2048. Right now, as I understand the spec, it actually allows for pretty much *anything* thanks to PKCS#7 support, without explicit requirement of supporting any particular set of supported DigestEncryptionAlgorithmIdentifier (RFC 2315).

Which leads me to think RSA 2048 is the "safe choice" for widespread compatibility, but that a vendor could ship support for even PQC algorithms *today*. 🤔

The deprecation only removed the ability to enumerate what the firmware can do, not its ability to do it.

> Another related "de-facto" standard is that the Secure Boot update Microsoft's been rolling out over the past several months is a 2048 bit RSA cert

~~I think it's not controversial to say Microsoft always had a strong say in UEFI ;)~~ I indeed looked at that. I think it goes with the above argument; Microsoft's docs also use the wording ["at least RSA 2048"](https://learn.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-secure-boot#:~:text=All%20firmware%20components%20must%20be%20signed%20using%20at%20least%20RSA%2D2048%20with%20SHA%2D256) but stuck with RSA 2048 for compatibility reasons.

> It's not uncommon for hardware vendors to only test against Microsoft OSes

There are other operating systems??? :o

Issue #624, opened 2025-11-27
Perform ZFS scrubbing


stgraber:
We should perform automatic weekly scrubs of our pools with an option to tweak the frequency in configuration. We should also add a `/1.0/storage/:scrub-pool` API route to trigger a manual scrub.

Pool status should reflect whether a scrub is ongoing.

luissimas:
Hey @stgraber, I'm interested in working on this feature. I still have to setup my development environment for IncusOS, but this seems like an interesting feature to get started :).

If you think it makes sense that I start working on this, I have a question: do we already have a strategy to perform periodic actions on IncusOS? I'm assuming we'd want to use systemd timers for this.

stgraber:
I've assigned it to you now.

I'll let @gibmat answer on the best approach for the timing logic. We can indeed use a systemd timer per pool for that, or we can have incus-osd handle those itself directly, similar to what's done for system updates.

gibmat:
Initially, I'd lean towards leveraging the systemd timers, since that would be one less thing we need to worry about in `incus-osd`. On my Debian systems the `zfs-linux` packages ship with such timers and that's how I have periodic scrubs of my zpools configured.

stgraber:
The manual scrub API route will be very easy to add, just the REST endpoint, a function in the zfs package and an extra entry in the CLI.

For the scheduled one. I guess the easy way to handle this would be expanding `api.SystemStorageConfig` to include a `ScrubSchedule` config. This would then default to something like `Sun 04:00` for a weekly 4am run on Sunday. Running sequentially on every pool.

@gibmat does that work? ^

stgraber:
Ah and we should make sure that we have the scrub status in the state data too, basically `State` should indicate that the pool is being scrubbed and we may want a `LastScrub` or similar field to show the last scrub result.

luissimas:
Thanks for the directions!

I'll start with the manual scrub API endpoint + status reporting and leave the scheduling for a separate PR. I'll start setting up my local environment today.

luissimas:
Now that #665 is merged, I'll start working on the scheduling part. I'll take a bit of time to get more familiar with systemd timers and share an implementation proposal here in the next couple of days.

I'm initially leaning towards writing a more generic `schedule` package to manage systemd timers that would be ZFS agnostic and could then be used for any other use case. We would then use that package to configure timers for each pool in response to changes in configuration and/or when creating a storage pool.

[...]

PR #665, opened 2025-12-04
Add ZFS scrub endpoint

luissimas:
# Description

This PR adds the capability of triggering and viewing the status of a scrub for a local ZFS storage pool via the `incus-osd` API. It also includes the CLI command to use this endpoint.

This is the first PR in a series to fulfill the feature described in #624. Performing automatic scrubs via scheduling is not in the scope of this PR, as I believe it's something that should be discussed and reviewed separately.

# Testing

I tested this implementation locally and also added an integration test in `incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py` asserting the basic scrub behavior.

# Example

Here's an example of usage demonstrating the scrub endpoint behavior as well as the `incus admin os system storage show` output.

```bash
root@devbox:~# incus query -X POST /os/1.0/system/storage/:scrub-pool -d '{"name": "local"}' --raw
{
        "type": "sync",
        "status": "Success",
        "status_code": 200,
        "operation": "",
        "error_code": 0,
        "error": "",
        "metadata": {}
}

root@devbox:~# incus query -X POST /os/1.0/system/storage/:scrub-pool -d '{"name": "local"}' --raw
{
        "type": "error",
        "status": "",
        "status_code": 0,
        "operation": "",
        "error_code": 409,
        "error": "scrub already in progress",
        "metadata": null
}

root@devbox:~# incus admin os system storage show
config: {}
state:
  drives:
  - boot: true
    bus: scsi
    capacity_in_bytes: 5.36870912e+10
    id: /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root
    member_pool: local
    model_family: QEMU
    model_name: QEMU HARDDISK
    remote: false
    removable: false
    serial_number: incus_root
  pools:
  - devices:
    - /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11
    encryption_key_status: available
    last_scrub:
      end_time: "2025-12-13T16:27:32Z"
      errors: 0
      examined_bytes: 3.70860032e+08
      start_time: "2025-12-13T16:27:29Z"
      state: FINISHED
      to_examine_bytes: 3.71183616e+08
    name: local
    pool_allocated_space_in_bytes: 3.71449856e+08
    raw_pool_size_in_bytes: 1.7716740096e+10
    state: ONLINE
    type: zfs-raid0
    usable_pool_size_in_bytes: 1.7716740096e+10
    volumes:
    - name: incus
      quota_in_bytes: 0
      usage_in_bytes: 3.69025024e+08
      use: incus
```

### DIFF
--- a/incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py
+++ b/incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+import time
 
 from .incus_test_vm import IncusTestVM, IncusOSException, util
 
@@ -390,3 +391,60 @@ def TestIncusOSAPISystemStorageLocalPoolRecoverFreshInstall(install_image):
             result = vm.APIRequest("/1.0/system/storage/:import-pool", method="POST", body="""{"name":"local","type":"zfs","encryption_key":""" + '"' + encryption_key + '"}')
             if result["status_code"] != 200:
                 raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
+
+
+def TestIncusOSAPISystemStorageLocalPoolScrub(install_image):
+    test_name = "incusos-api-system-storage-local-pool-scrub"
+    test_seed = {
+        "install.json": """{"target":{"id":"scsi-0QEMU_QEMU_HARDDISK_incus_root"}}""",
+    }
+
+    test_image, incusos_version = util._prepare_test_image(install_image, test_seed)
+
+    with IncusTestVM(test_name, test_image) as vm:
+        vm.WaitSystemReady(incusos_version)
+
+        # Get current storage state.
+        result = vm.APIRequest("/1.0/system/storage")
+        if result["status_code"] != 200:
+            raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
+
+        if len(result["metadata"]["state"]["pools"]) != 1:
+            raise IncusOSException("expected exactly one pool")
+
+        if "last_scrub" in result["metadata"]["state"]["pools"][0]:
+            raise IncusOSException("expected no last_scrub to be reported since to scrub was requested")
+
+        # Scrub the pool.
+        result = vm.APIRequest("/1.0/system/storage/:scrub-pool", method="POST", body="""{"name":"local"}""")
+        if result["status_code"] != 200:
+            raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
+
+        # Wait for scrub to complete.
+        time.sleep(5)
+
+        # Get current storage state.
+        result = vm.APIRequest("/1.0/system/storage")
+        if result["status_code"] != 200:
+            raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
+
+        if len(result["metadata"]["state"]["pools"]) != 1:
+            raise IncusOSException("expected exactly one pool")
+
+        if "last_scrub" not in result["metadata"]["state"]["pools"][0]:
+            raise IncusOSException("expected last_scrub to be reported after scrubbing the pool")
+
+        if "start_time" not in result["metadata"]["state"]["pools"][0]["last_scrub"]:
+            raise IncusOSException("expected start time to be reported on the last scrub")
+
+        if "end_time" not in result["metadata"]["state"]["pools"][0]["last_scrub"]:
+            raise IncusOSException("expected end time to be reported on the last scrub")
+
+        if result["metadata"]["state"]["pools"][0]["last_scrub"]["state"] != "FINISHED":
+            raise IncusOSException("expected last scrub to have 'FINISHED' status")
+
+        if result["metadata"]["state"]["pools"][0]["last_scrub"]["progress"] != "100.00%":
+            raise IncusOSException("expected progress to be reported on the last scrub")
+
+        if result["metadata"]["state"]["pools"][0]["last_scrub"]["errors"] != 0:
+            raise IncusOSException("expected 0 errors to be reported on the last scrub")


doc/rest-api.yaml
luissimas, line 1: Not really sure what happened on this one, I just ran `make update-api` but got a diff bigger than I expected. Should I clean up this and ensure that the only changes are related to the new endpoint?

incus-osd/cli/cli_system.go
```diff
@@ -180,7 +180,16 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
...
+				// Scrub pool.
+				scrubPoolCmd := cmdGenericRun{
+					os:          c.os,
+					action:      "scrub-pool",
+					description: "Scrub a local ZFS pool",
+					endpoint:    "system/storage",
+					hasData:     true,
+				}
+
+				return []*cobra.Command{createVolumeCmd.command(), deletePoolCmd.command(), deleteVolumeCmd.command(), importPoolCmd.command(), wipeDriveCmd.command(), scrubPoolCmd.command()}
```
luissimas, line 192: Is this really all it takes to add a new command? This will then be included in the `incus admin os` CLI, right? How do you normally test these kind of changes?

incus-osd/internal/zfs/zfs.go
```diff
@@ -810,3 +810,21 @@ func DestroyDataset(ctx context.Context, poolName string, name string, force boo
...
 }
+
+func ScrubZpool(ctx context.Context, poolName string) error {
+	// Check if the zpool exists.
+	if !storage.PoolExists(ctx, poolName) {
+		return errors.New("zpool '" + poolName + "' doesn't exist")
+	}
+
+	// TODO: check scrub status. We should be able to detect that a scrub is already
+	// in progress and provide a better error message to the user.
```
luissimas: The goal here is to provide and error message better than the default one that ZFS will throw if we try to scrub a pool that's already being scrubbed. I think we should:

1. Inform the user that the pool is already being scrubbed
2. Guide the user by showing the command that shows the status of the storage pool

gibmat: A simple error like `"scrub already in progress for " + poolName` should be sufficient. In the code that's actually performing the scrub we don't know where the function might be called from, so we shouldn't try to give any hints in the error.

doc/rest-api.yaml
gibmat: No, there's no need to try to fix up the output since it's non-deterministic in its ordering.

incus-osd/cli/cli_system.go
gibmat: Yeah, @stgraber setup a pretty easy way to add commands to the `cli` package that's consumed by Incus when it builds the client executable.

The IncusOS tests operate directly with the REST API to test correctness. Testing the new cli command can be done pretty easily by locally building the Incus client so it picks up your local changes in the IncusOS `cli` package.

incus-osd/cli/cli_system.go
```diff
@@ -180,7 +180,16 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 					confirm:     "wipe the drive",
 				}
 
-				return []*cobra.Command{createVolumeCmd.command(), deletePoolCmd.command(), deleteVolumeCmd.command(), importPoolCmd.command(), wipeDriveCmd.command()}
+				// Scrub pool.
+				scrubPoolCmd := cmdGenericRun{
+					os:          c.os,
+					action:      "scrub-pool",
+					description: "Scrub a local ZFS pool",
```
stgraber: We usually try to avoid mentioning ZFS as theoretically the storage pool API could support other storage types (LVM being the most likely to be made available through that).

gibmat:
> Reporting the ZFS scrub progress in the pool status is something that I think should be in the scope of this PR, as it will enable `incus-osd` to check if a scrub is already in progress before triggering a new one when the API is called. I haven't got to that yet, but I plan on including it in this PR.

Yeah, it should be easy enough to use the existing `State` field in the `SystemStoragePool` API struct to report when a pool is performing a scrub.

> I'm assuming that for testing this one we'd want a new test in `incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py`, right?

Yes, we're trying to make sure we add tests for any new/changed REST APIs to ensure things work as expected and to (hopefully) make it easier to detect any regressions.

incus-osd/internal/rest/api_system_storage.go
```diff
@@ -533,3 +533,70 @@ func (*Server) apiSystemStorageDeleteVolume(w http.ResponseWriter, r *http.Reque
 
 	_ = response.EmptySyncResponse.Render(w)
 }
+
+// swagger:operation POST /1.0/system/storage/:scrub-pool system system_post_storage_scrub_pool
+//
+//	Scrub local pool
+//
+//	Starts a ZFS scrub for a local storage pool.
```
stgraber: Same as above, the description should be generic and not mention ZFS specifically here

stgraber (changes requested):
Structure looks good, just a couple of tweaks to wording in CLI and API description

stgraber:
Ah, looks like you're missing a comment/doc string on the new ZFS function too

incus-osd/cli/cli_system.go
luissimas: Nice! I'll build the CLI locally to test the command E2E.

incus-osd/internal/zfs/zfs.go
luissimas: Ok, that makes a lot of sense

luissimas:
Thanks a lot for the input here! @stgraber I made the wording adjustments to not mention ZFS directly. I'll go ahead and start working on the status reporting, and leveraging it in the test suite.

luissimas:
Made some progress with status reporting. The last bit missing here is checking the scrub status before trying to scrub the pool manually. Should be ready to review in the next few days.

incus-osd/api/system_storage.go
```diff
@@ -50,6 +53,25 @@ type SystemStoragePoolVolume struct {
...
+
+// SystemStoragePoolScrubStatus represents the status of a scrub in a pool.
+type SystemStoragePoolScrubStatus struct {
+	State          SystemStoragePoolScrubState `json:"state"`
+	StartTime      time.Time                   `json:"start_time"`
+	EndTime        time.Time                   `json:"end_time"`
+	ToExamineBytes int                         `json:"to_examine_bytes"`
+	ExaminedBytes  int                         `json:"examined_bytes"`
+	Errors         int                         `json:"errors"`
+}
```
luissimas, line 75: I've tried to keep the fields here ZFS agnostic, but I'm not familiar with how LVM deals with scrubs. I assume that any information here should be common to both implementations.

stgraber: Maybe replace ToExamineBytes and ExamineBytes by a Progress string?

That would allow reporting a more user friendly percentage value and would be more flexible with any other scrub implementation we may need to deal with in the future.

incus-osd/internal/storage/storage.go
```diff
@@ -30,9 +31,46 @@ type LsblkOutput struct {
...
+	ZpoolFinished ZpoolScrubState = "FINISHED"
+)
+
+var zpoolToScrubStateMap = map[ZpoolScrubState]api.SystemStoragePoolScrubState{
+	ZpoolScanning: api.ScrubInProgress,
+	ZpoolFinished: api.ScrubFinished,
+}
+
+func zpoolScrubStateToPoolScrubState(state ZpoolScrubState) api.SystemStoragePoolScrubState {
+	if mapped, ok := zpoolToScrubStateMap[state]; ok {
```
stgraber: We prefer not to do declaration + check in our code, so instead split this into two lines:
```
mapped, ok := zpoolToScrubStateMap[state]
if ok {
```

incus-osd/cli/cli_system.go
```diff
@@ -180,7 +180,16 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 					confirm:     "wipe the drive",
 				}
 
-				return []*cobra.Command{createVolumeCmd.command(), deletePoolCmd.command(), deleteVolumeCmd.command(), importPoolCmd.command(), wipeDriveCmd.command()}
+				// Scrub pool.
+				scrubPoolCmd := cmdGenericRun{
+					os:          c.os,
+					action:      "scrub-pool",
+					description: "Scrub a local pool",
```
stgraber: I'd just make it `Scrub the storage pool` as a pool can be built on top of a remote NVME-over-TCP or iSCSI volume.

stgraber:
Looks great, just a few minor details to tweak!

incus-osd/api/system_storage.go
luissimas: I think it makes sense to have a `Progress` information instead of `ToExamineBytes` and `ExamineBytes`.

Do you think this value should be a 0-100 percentage value already formatted for user display (e.g. "70%")? Initially I'd lean towards just formatting it on the client and keeping a 0-1 float value on the API, but I think this would introduce more complexity for no added value right now.

stgraber: The client right now is very very stupid so doesn't really do any processing and we don't really know what kind of progress data we could get on LVM or btrfs or anything else. So having a string which can be rendered server-side based on the kind of pool should make that easier.

luissimas: I've changed it to report the progress as a percentage string in a `Progress` field. Had to deal with some edge cases in displaying since ZFS works with some estimates.

I think the logic for properly formatting the progress percentage is slightly more complex than what I expected, so I'm not 100% sure that we want it in the server-side. Please let me know what you think.

stgraber (approved):

incus-osd/api/system_storage.go
stgraber: Code is reasonably simple and looks good to me, fine to have server side.

stgraber:
Looks good!

PR #667, opened 2025-12-04
Add swtpm fallback for systems without physical TPM

### LINKED ISSUE #370: Support swtpm for systems that can't have a physical TPM 2.0 module
It should be possible to run a software TPM (swtpm) using a partition for its state.
This won't be as secure as swtpm could be attacked in theory from a compromised OS but it would still unblock installing on some systems, particularly consumer grade Arm systems.

Doing this should come with a persistent message in the TUI and security endpoint so it's easy to detect such systems, allowing for operations center or other management tools to refuse to trust them.

gibmat:
This PR adds the ability for IncusOS to run on systems without a physical TPM by relying on `swtpm`. Notably, while this will still provide some level of protection for LUKS-encrypted partitions, there will by necessity be some compromises due to the TPM state being easily accessible.

Manually tested with `swtpm`:
* Running IncusOS in "live" mode
* Adding/changing/removing recovery passphrase
* SecureBoot `db` and `dbx` certificate changes
* SecureBoot UKI signing key rotation
* OS and application updates

Closes #370

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -25,6 +25,116 @@ definitions:
         title: AccessEntry represents an entity having access to the resource.
         type: object
         x-go-package: github.com/lxc/incus/v6/shared/api
+    Artifact:
+        properties:
+            architectures:
+                description: Architectures used to match VMs to an artifact.
+                example: x86_64
+                items:
+                    type: string
+                type: array
+                x-go-name: Architectures
+            description:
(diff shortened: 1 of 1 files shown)


gibmat:
@stgraber This should be ready for review.

stgraber (approved):

stgraber:
@gibmat needs a rebase

PR #679, opened 2025-12-09
app-build/tailscale: fix feature flags to match latest version

quagsirus:
Permanent solution to #561

Removing `ts_omit_logtail` was a temporary fix, no longer needed as-of Tailscale v1.92.0 / https://github.com/tailscale/tailscale/commit/8aa46a395631f49b98104eea2b13432f1a196375


quagsirus:
Check looks like it succeeded despite being marked as cancelled? May be worth making the timeout a bit more lenient than 1h30m considering that job averages around 1h25m

PR #683, opened 2025-12-09
incus-osd/systemd/networkd: Only trigger network devices

stgraber:


stgraber:
We very much don't want to trigger storage device rescan on network config changes, when said storage devices are accessed over the currently broken network ;)

gibmat (approved):

stgraber:
@gibmat extra fix for a bug I ran into during validation ^

gibmat (approved):

PR #691, opened 2025-12-11
incus-osd/systemd/networkd: Don't delete bridges that are in use

### LINKED ISSUE #687: Modifying the network config (like adding a VLAN to vlan_tags) break all instances
When modifying `vlan_tags` on the IncusOS system, all VMs that are attached to the physical network stop responding and require a reboot to get back on the network.

Here we can see the before and after of the interfaces on the IncusOS system:
```
bash-5.2# ip link | grep master.*enp6s0
19: _p10666a9d8da3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master enp6s0 state UP mode DEFAULT group default qlen 1000
33: _i10666a9d8da3@_venp6s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master enp6s0 state UP mode DEFAULT group default qlen 1000
36: tapd0afdbf8: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master enp6s0 state UP mode DEFAULT group default qlen 1000
37: tap05541ea3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master enp6s0 state UP mode DEFAULT group default qlen 1000

bash-5.2# ip link | grep master.*enp6s0
19: _p10666a9d8da3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq master enp6s0 state UP mode DEFAULT group default qlen 1000
38: _i10666a9d8da3@_venp6s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master enp6s0 state UP mode DEFAULT group default qlen 1000
bash-5.2#
```

I suspect it's our logic to delete any affected network interface to have it fully reconfigured which is backfiring here. We probably need to make an exception on it for any bridge which contains external (not created by IncusOS) interfaces.

stgraber:
Closes #687

### DIFF
--- a/incus-osd/internal/systemd/networkd.go
+++ b/incus-osd/internal/systemd/networkd.go
@@ -1391,6 +1391,8 @@ func generateLinkSectionContents(addresses []string, requiredForOnline string) s
 }
 
 func cleanupStaleDevices(ctx context.Context, oldCfg *api.SystemNetworkConfig, newCfg *api.SystemNetworkConfig) error {
+	deleteInterfaces := []string{}
+
 	// Check for changed/deleted interfaces.
 	for oldIndex := range oldCfg.Interfaces {
 		newIndex := slices.IndexFunc(newCfg.Interfaces, func(i api.SystemNetworkInterface) bool {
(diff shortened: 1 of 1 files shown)


incus-osd/internal/systemd/networkd.go
```diff
@@ -1478,15 +1488,50 @@ func cleanupStaleDevices(ctx context.Context, oldCfg *api.SystemNetworkConfig, n
...
+		deleteNetworkDevice(ctx, deleteInterfaces...)
+	}
+
 	return nil
 }
 
+// isBridgeInUse checks if a bridge exists and only contains internal ports.
+func isBridgeInUse(device string) bool {
+	entries, err := os.ReadDir(filepath.Join("/sys/class/net", device))
+	if err != nil && errors.Is(err, os.ErrNotExist) {
```
gibmat: Do we want to handle other errors beside `ErrNotExist`? If so, it would probably be good for this function to return `(bool, error)`.

stgraber: Changed it to return false on all errors at this stage.

PR #720, opened 2025-12-20
Add support for WireGuard interfaces

### LINKED ISSUE #664: Add support for tunnel interfaces (wireguard) in network configuration
We should add support for basic tunnel interfaces in the network configuration.

Those shouldn't have a role as they'll effectively be management-only and won't be suitable to directly put instances on.

Initially we should just do wireguard but keep in mind that GRE, VXLAN and SIT tunnels would also be good candidates to add.

defect-track:
Adds support for Wireguard using systemd-networkd configuration files.

Closes #664

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -102,6 +102,7 @@ VLANs
 VMware
 VPN
 vSphere
+WireGuard
 WWN
 YAML
 Zabbly


stgraber:
@defect-track could you edit your commit message to include the Signed-off-by line?
There are also a few static analysis failures that should be looked at :)

defect-track:
@stgraber updated the code based on reported errors. For the documentation error we properly need to add WireGuard as this is an unknown word for the spellchecker as it looks. Added the --signoff option to the latest commit.

Will wait on the next run to see if there are more changes required.

doc/reference/system/network.md
```diff
@@ -154,6 +156,50 @@ Configure a VLAN with ID 123 on top of an active-backup bond composed of two int
...
+        "port": 51820,
+        "private_key": "AE1SCwtkp8ruDYlUa9x9wsoTzEOePl3P9sMdFFa9PmI=",
+        "addresses": [
+          "10.234.234.100/24",
+          "fd42:3cfb:8972:abcd::100/64"
+        ],
+        "routes" : [
+          "10.234.110.0/24",
+          "10.234.234.110/24",
+        ],
```
stgraber, line 180: That looks incorrect, no?

`routes` is marked as being a `SystemNetworkRoute` which would mean structs with `to` and `via` rather than just a list of strings.

incus-osd/api/system_network.go
```diff
@@ -88,6 +89,29 @@ type SystemNetworkFirewallRule struct {
...
+	FirewallRules     []SystemNetworkFirewallRule  `json:"firewall_rules,omitempty"      yaml:"firewall_rules,omitempty"`
+	Name              string                       `json:"name"                          yaml:"name"`
+	MTU               int                          `json:"mtu,omitempty"                 yaml:"mtu,omitempty"`
+	Peers             []SystemNetworkWireguardPeer `json:"peers,omitempty"               yaml:"peers,omitempty"`
+	Port              int                          `json:"port,omitempty"                yaml:"port,omitempty"`
+	PrivateKey        string                       `json:"private_key,omitempty"         yaml:"private_key,omitempty"`
+	RequiredForOnline string                       `json:"required_for_online,omitempty" yaml:"required_for_online,omitempty"`
+	Roles             []string                     `json:"roles,omitempty"               yaml:"roles,omitempty"`
+	Routes            []SystemNetworkRoute         `json:"routes,omitempty"              yaml:"routes,omitempty"`
+}
```
stgraber, line 115: Tiny nit but we usually try to keep those fields sorted alphabetically, this doesn't appear to be the case here (MTU/Name).

doc/reference/system/network.md
defect-track: Next to missing `to` and `via` it has to go `via` an IP address and not a network.

incus-osd/api/system_network.go
defect-track: Looks like my alphabet was off a bit.

stgraber:
I'm waiting for #724 to clear, then I'll look into this one, will rebase, make any small tweak needed and send it to get merged, thanks!

doc/reference/system/network.md
```diff
@@ -154,6 +156,52 @@ Configure a VLAN with ID 123 on top of an active-backup bond composed of two int
 }
 ```
 
+#### WireGuard
+
+Configure a WireGuard interface with two peers (providing a private_key is optional and will be created if empty):
+
+```
+{
```
gibmat, line 164: Need to wrap this in a `{"config": ... }` block. For the examples, someone should be able to copy-n-paste the example directly into the API and it should work (assuming correct values, etc).

incus-osd/internal/systemd/networkd.go
```diff
@@ -1660,3 +1992,20 @@ func mangleUSBNICs(config *api.SystemNetworkConfig) {
 		}
 	}
 }
+
+func checkWireguardPrivateKeys(ctx context.Context, networkCfg *api.SystemNetworkConfig) error {
+	// check for private key for each Wireguard.
```
gibmat: Minor nit: We try to make comments proper sentences: capitalized and ending with a period.

incus-osd/internal/systemd/networkd.go
```diff
@@ -360,6 +392,200 @@ func UpdateNetworkState(ctx context.Context, n *api.SystemNetwork) error {
 	return nil
 }
 
+// getWireguardState runs various commands to gather wireguard state for a specific wireguard interface.
+func getWireguardState(ctx context.Context, ifaceType string, iface string) (api.SystemNetworkInterfaceState, error) {
```
gibmat: `ifaceType` seems redundant for this function?

gibmat (changes requested):

stgraber:
I've rebased the branch, merged the two commits together and addressed the review comments from @gibmat.

I still need to do a review myself, split the commits a bit, but we should at least have it at a stage now where the tests should all be passing.

stgraber (approved):

PR #722, opened 2025-12-20
Tweak IPv6 handling slightly

stgraber:


stgraber:
The motivation for this one is IncusOS very frequently reporting an IPv4 address to Operations Center when it's configured to require the IPv6 family and the IPv4 address isn't functional for Operations Center communication.

So I keep having to bounce servers in the hope of avoiding that race and having them correctly register with their IPv6 address...

Issue #638, opened 2025-11-29
Add wake on lan option into network interface configuration


cfouche3005:
Basically the title of the issue.
Even if wake on lan is enable in the bios, wake on lan still need to be enabled in the OS, generally, via ethtool

stgraber:
We should just expose the following networkd link properties:
 - `WakeOnLan`
 - `WakeOnLanPassword`

They'll only make sense on physical NICs and should be ignored everywhere else.

Triiltz:
I was taking a quick look here and had some thoughts.

1. Add `wake_on_lan` and `wake_on_lan_password` fields to both `SystemNetworkInterface` and `SystemNetworkBond` structs in `api/system_network. go`

2.  Modify `generateLinkFileContents()` in `networkd. go` to include the `WakeOnLan=` directive in the [Link] section of generated . link files, and this would be an optional, right?

3. Add validation to ensure valid WakeOnLan values (off, phy, unicast, multicast, broadcast, arp, magic, secureon) and that password is only used with secureon. Just a thought, not sure if it matches what you have in mind here

Does this make sense with what you have in mind for this?

Also, should WoL be supported on bonds as well, or just on individual interfaces?

stgraber:
Yeah, that sounds good to me. I don't know how relevant wake-on-lan would be on a bond interface as depending on the mode, I wouldn't expect the switch to forward those at all unless bonding is properly negotiated first. Machines using a bond are also somewhat more likely to have a proper BMC of some kind and so not have to rely on wake-on-lan to turn themselves on.

So it may make sense to just add the two options to interfaces for now and only expand it to bonds when someone actually runs into that situation and can confirm a functional wake-on-lan environment combined with bonding.

PR #724, opened 2025-12-21
Add ethernet settings on physical network interfaces

### LINKED ISSUE #723: Add support for disabling tcp segmentation offloading for an interface
I'm attempting to run incusos on a trio of 8th gen Intel NUCs, however at the moment they'll drop off the network seemingly randomly due to a hw issue exasperated by the e1000e Linux driver. Near as I can tell the only way to fix this is to use ethtool to turn off tso, but I'm not finding any way to do that in incusos.

AmandaCameron:
Fixes #723 -- I'm not entirely happy with the option name I chose, but I'm not great at naming things, so if there's a better name do let me know. Passes `go test` locally, but didn't try in a VM yet, need to get mkosi installed on my laptop first.

### DIFF
--- a/incus-osd/internal/systemd/networkd_test.go
+++ b/incus-osd/internal/systemd/networkd_test.go
@@ -147,6 +147,7 @@ interfaces:
     required_for_online: both
     hwaddr: AA:BB:CC:DD:EE:01
     ethernet:
+      disable_energy_efficient: true
       disable_ipv4_tso: true
       disable_ipv6_tso: true
       disable_gro: true
@@ -170,6 +171,7 @@ bonds:
       - AA:BB:CC:DD:EE:02
       - AA:BB:CC:DD:EE:03
     ethernet:
+      disable_energy_efficient: true
       disable_ipv4_tso: true
       disable_ipv6_tso: true
 `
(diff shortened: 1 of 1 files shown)


AmandaCameron:
Testing in a local VM is turning out to be more complicated than I expected. Seems `make test` assumes incus is running locally, when I've only got it available remotely, or via a libvirt instance of incusos

stgraber:
I don't think we want to get an extra entry in that struct for every type of offload, so having a sub-struct for it is probably best.

Something like:
```
type SystemNetworkOffloading struct {
    DisableGRO bool `json:"disable_gro,omitempty" yaml:"disable_gro,omitempty"`
    DisableGSO bool `json:"disable_gso,omitempty" yaml:"disable_gso,omitempty"`
    DisableIPv4TSO bool `json:"disable_ipv4_tso,omitempty" yaml:"disable_ipv4_tso,omitempty"`
    DisableIPv6TSO bool `json:"disable_ipv6_tso,omitempty" yaml:"disable_ipv6_tso,omitempty"`
}
```

For now they're all "disable" flags for offloads that should be there, but that could also start getting positive values for offloading that's typically disabled.

stgraber:
Then both `SystemNetworkInterface` and `SystemNetworkBond` would need to get a `Offloading SystemNetworkOffloading` struct field to contain those flags.

AmandaCameron:
Implemented separate switches for the ones you listed.

I went with making them just be unmentioned if they're not set to disabled, because I didn't want to risk the code causing them to be turned on when the kernel has them off for some reason, causing potential problems.

AmandaCameron:
ah, forgot I switched my git email, will fix the DCO now

stgraber:
Thanks, I'll take a look, tweak a bit and land it!

Other ethtool flags were brought up on the forum today (energy efficient ethernet) so I'm going to check if that makes sense to still fit under Offloading or if we need an even more generic thing like `Ethernet` or something instead for those flags.

stgraber:
@gibmat given my additions on this one, can you give this a quick review?

cfouche3005:
Does the WOL settings (https://github.com/lxc/incus-os/issues/638) also related to this ? Should they also be add to this PR ?

stgraber:
They'll most likely get added to that struct. I was going to take a look at that next once this one is in with the general infrastructure for this.

incus-osd/api/system_network.go
```diff
@@ -39,6 +39,7 @@ type SystemNetworkConfig struct {
 // SystemNetworkInterface contains information about a network interface.
 type SystemNetworkInterface struct {
 	Addresses         []string                    `json:"addresses,omitempty"           yaml:"addresses,omitempty"`
+	Ethernet          SystemNetworkEthernet       `json:"ethernet"                      yaml:"ethernet"`
```
gibmat: Mark these two new fields as `omitempty` and make a pointer so they match the other fields and don't clutter up the API response if not set?

incus-osd/internal/systemd/networkd.go
```diff
@@ -894,6 +894,35 @@ func waitForSystemdTimesyncd(ctx context.Context, timeout time.Duration) error {
 func generateLinkFileContents(networkCfg api.SystemNetworkConfig) []networkdConfigFile {
 	ret := []networkdConfigFile{}
 
+	generateEthernet := func(s api.SystemNetworkEthernet) string {
```
gibmat: This function is now generating two different sections of the configuration file -- it might be less error-prone to split it into two functions: `generateLinkSection()` and `generateEnergyEfficientEthernet()`.

stgraber: I considered that but it felt weird having multiple functions generate text output for the same input struct so I went with the single function.

This will get further expanded with the wake-on-lan stuff later today.

Down the line, I think we should switch all the networkd stuff to Go templates but I didn't feel like doing that change as part of this :)

incus-osd/api/system_network.go
stgraber: Fixed!

incus-osd/internal/systemd/networkd.go
stgraber: https://github.com/lxc/incus-os/issues/728

stgraber:
@gibmat should be good to go. I updated the API and filed an issue so we can keep track switching to text/template in the future.

AmandaCameron:
Will this require waiting for incus 1.21 or will this just work when the stable incusos branch is next updated? I ask because this being a blocker is exactly what motivated me to make the PR in the first place. (Buggy e1000e hw <-> driver interactions)

stgraber:
@AmandaCameron it will be in the next testing and then stable IncusOS image.
We have a few more things landing right now but there's a good chance I'll tag a testing image later today or tomorrow and we'll be pushing something to stable this week.

AmandaCameron:
Got it, hopefully I can switch my remaining two incus nodes to incusos sometime soon then. Having to physically go and detach/reatatch the ethernet cable on the single NUC i did as a canary is a bit tedius, and halted my migration entirely.

PR #727, opened 2025-12-22
Secure Boot certificate updates

gibmat:
Bit of a random grab bag of Secure Boot certificate related updates:

* Properly calculate offset when parsing .auth files. This should allow us to handle a variety of supported signature types rather than our previously hard coded assumption of RSA+SHA256.
* Don't assume buffer size will be the same when updating EFIVariableAuthority
* Update test certificate generation to more closely match production certificates
* Fix commands to test Secure Boot updates in github workflow
* Update docs to better reflect our Secure Boot hierarchy

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -1,4 +1,5 @@
 backend
+CAs
 CDN
 CDROM
 Ceph


stgraber:
@gibmat second time failing at the same spot, is tat an actual failure?

gibmat:
Yeah, something odd is happening; I'll take a look at it in the morning.

gibmat:
Figured it out -- another edge case revealed by changing up the test certificates.

PR #732, opened 2025-12-24
Adds initrd modules to support PERC H965i

gdonval:
Adds module `mpi3mr` and its dependency, `scsci_transport_sas`, to support PERC H965i cards (used by Dell PowerEdge).

Signed-off-by: Gaël Donval <github.com@umessage.me>


stgraber:
I'd expect the dependency to get pulled in on its own, but doesn't hurt to have it explicitly :)

Thanks!

PR #736, opened 2025-12-26
Expand SMART support

stgraber:


incus-osd/api/system_storage.go
```diff
@@ -94,6 +94,14 @@ type SystemStorageDrive struct {
...
 	Enabled bool `json:"enabled" yaml:"enabled"`
 	Passed  bool `json:"passed"  yaml:"passed"`
+
+	PowerOnHours     int `json:"power_on_hours,omitempty"      yaml:"power_on_hours,omitempty"`
+	DataUnitsRead    int `json:"data_units_read,omitempty"     yaml:"data_units_read,omitempty"`
+	DataUnitsWritten int `json:"data_units_written,omitempty"  yaml:"data_units_written,omitempty"`
+	AvailableSpare   int `json:"available_spare,omitempty"     yaml:"available_spare,omitempty"`
+	PercentageUsed   int `json:"percentage_used,omitempty"     yaml:"percentage_used,omitempty"`
+	RawReadErrorRate int `json:"raw_read_error_rate,omitempty" yaml:"raw_read_error_rate,omitempty"`
+	SeekErrorRate    int `json:"seek_error_rate,omitempty"     yaml:"seek_error_rate,omitempty"`
```
gibmat: Would we want to also include `Reallocated_Event_Count`?

gibmat (approved):

incus-osd/api/system_storage.go
stgraber: Sounds useful. I was just going from the list from the forum, but having the re-allocation count would probably be good.

stgraber: Actually, Reallocated_Sector_Ct seems to be the one that's present on both HDD and SSD and is more specific than the event count one.

stgraber: Done, added it now!

PR #738, opened 2025-12-28
incus-osd/state: Fix json tag in SuccessfulBoot

luissimas:
Noticed what I believe is a typo in this JSON struct tag while working in another feature.


stgraber (approved):

stgraber:
Good catch!

PR #739, opened 2025-12-28
Add periodic scrub scheduling for storage pools

### LINKED ISSUE #624: Perform ZFS scrubbing

luissimas:
# Description

Closes #624. This PR adds the support for performing automatic periodic scrubbing of storage pools (currently only ZFS). The scrub is performed sequentially (instead of in parallel) on every pool to avoid high resource usage on systems with many storage pools.

Users can control the scrub schedule by setting the `scrub_schedule` config option in the storage config. This is a cron string that determines the frequency of the scrubbing.

The implementation is mostly complete. The only thing that is missing is returning the `scrub_schedule` value from the API (useful for seeing it in `incus admin os system storage show` and to populate the edit buffer for `incus admin os system storage edit`). I assume this would require us to return the stored OS state in the API, rather than just fetching the state of the ZFS pools using `zfs` commands.

# Testing

The testing I performed on this one was mostly manual. I don't currently see a good way of testing this automatically in a way that wouldn't hurt the CI execution speed, but I'm open to suggestions in this regard.

### DIFF
--- a/incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py
+++ b/incus-osd/tests/incusos_tests/tests_incusos_api_system_storage_local_pool.py
@@ -93,7 +93,7 @@ def TestIncusOSAPISystemStorageLocalPoolExpandRAID0(install_image):
                 raise IncusOSException("pool has unexpected name: " + result["metadata"]["state"]["pools"][0]["name"])
 
             # Extend the "local" pool with the second drive
-            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"pools":[{"name":"local","type":"zfs-raid0","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1"]}]}}""")
+            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"scrub_schedule": "0 4 * * 0", "pools":[{"name":"local","type":"zfs-raid0","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1"]}]}}""")
             if result["status_code"] != 200:
                 raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
 
@@ -136,7 +136,7 @@ def TestIncusOSAPISystemStorageLocalPoolExpandRAID0(install_image):
                 raise IncusOSException("pool has unexpected name: " + result["metadata"]["state"]["pools"][0]["name"])
 
             # Don't allow removal of the main system partition from the pool
-            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"pools":[{"name":"local","type":"zfs-raid0","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1",""]}]}}""")
+            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"scrub_schedule": "0 4 * * 0", "pools":[{"name":"local","type":"zfs-raid0","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1",""]}]}}""")
             if result["status_code"] == 200:
                 raise IncusOSException("unexpected success removing main storage partition")
 
@@ -144,7 +144,7 @@ def TestIncusOSAPISystemStorageLocalPoolExpandRAID0(install_image):
                 raise IncusOSException("unexpected error message: " + result["error"])
 
             # Remove the second drive from the pool
-            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"pools":[{"name":"local","type":"zfs-raid0","devices":["","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11"]}]}}""")
+            result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"scrub_schedule": "0 4 * * 0", "pools":[{"name":"local","type":"zfs-raid0","devices":["","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11"]}]}}""")
             if result["status_code"] != 200:
                 raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
 
@@ -244,7 +244,7 @@ def TestIncusOSAPISystemStorageLocalPoolExpandRAID1(install_image):
                     raise IncusOSException("pool has unexpected name: " + result["metadata"]["state"]["pools"][0]["name"])
 
                 # Extend the "local" pool with the second drive
-                result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"pools":[{"name":"local","type":"zfs-raid1","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1"]}]}}""")
+                result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"scrub_schedule": "0 4 * * 0", "pools":[{"name":"local","type":"zfs-raid1","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1"]}]}}""")
                 if result["status_code"] != 200:
                     raise IncusOSException("unexpected status code %d: %s" % (result["status_code"], result["error"]))
 
@@ -293,7 +293,7 @@ def TestIncusOSAPISystemStorageLocalPoolExpandRAID1(install_image):
                     raise IncusOSException("pool has unexpected name: " + result["metadata"]["state"]["pools"][0]["name"])
 
                 # Can't add a third device to the "local" pool
-                result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"pools":[{"name":"local","type":"zfs-raid1","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk2"]}]}}""")
+                result = vm.APIRequest("/1.0/system/storage", method="PUT", body="""{"config":{"scrub_schedule": "0 4 * * 0", "pools":[{"name":"local","type":"zfs-raid1","devices":["/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk1-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11","/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk2"]}]}}""")
                 if result["status_code"] == 200:
                     raise IncusOSException("unexpected success adding third drive")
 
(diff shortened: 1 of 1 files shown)


incus-osd/internal/state/struct.go
```diff
@@ -164,3 +165,14 @@ func (s *State) ManagementAddress() net.IP {
 
 	return nil
 }
+
```
stgraber: Instead of a function figuring out the default, it's probably best to set the default value in `state/file.go` (`initialize`) and then have a state upgrade function which sets it for existing installations too.

incus-osd/internal/state/struct.go
```diff
@@ -80,6 +80,7 @@ type State struct {
 		Provider api.SystemProvider `json:"provider"`
 		Security api.SystemSecurity `json:"security"`
 		Update   api.SystemUpdate   `json:"update"`
+		Storage  api.SystemStorage  `json:"storage"`
```
stgraber, line 71: This may cause us to persist data that we don't want/need to persist, specifically the `Pools` part of `api.SystemStorage.Config`.

To avoid that, we'll need to tag the `Pools` field with `incusos:"-"`.

stgraber:
Looking pretty good to me!

Just caught a couple of small details, but overall, looks clean with the new scheduler.

incus-osd/internal/state/struct.go
```diff
@@ -24,6 +27,14 @@ type OS struct {
 	SuccessfulBoot bool   `jsno:"successful_boot"`
 }
 
+// JobName represents the name of a periodic job.
+type JobName string
```
gibmat: I think it would probably be a good idea to add the constants and logic for job scheduling to a new internal package. That would allow it to encapsulate details about the implementation (such as `gocron.Scheduler`) as well as make it easier to add additional jobs down the road.

incus-osd/internal/state/struct.go
```diff
@@ -35,6 +46,10 @@ type State struct {
 
 	UpdateMutex sync.Mutex `json:"-"`
 
+	// Mapping of job names to job UUIDs.
+	Jobs         map[JobName]uuid.UUID `json:"-"`
```
gibmat: Is there any value to writing these two fields to disk via the state struct? It looks like on startup `incus-osd` will always overwrite everything by calling `registerJobs()`. Plus, the `gocron.Scheduler` struct looks like it doesn't have any exported fields and I'm not sure if our current serialization logic would be happy with that.

If we only need an in-memory state to use for REST API calls updating the schedule, also tag these two fields with `incusos:"-"`.

incus-osd/internal/state/struct.go
```diff
@@ -164,3 +165,14 @@ func (s *State) ManagementAddress() net.IP {
 
 	return nil
 }
+
+// ScrubSchedule returns the configured schedule for scrubbing storage pools.
+func (s *State) ScrubSchedule() string {
```
gibmat: I'd move this to the `api.SystemStorage` struct, so it can be used in a more generic fashion.

Also, a minor nit but if you checkout this specific commit in the branch, it won't compile because the `ScrubSchedule` field isn't added for a couple more commits. Not too big of an issue in the context of the PR, but if someone is trying to git bisect an error sometime, they might end up on this commit and encounter an unrelated compile error.

incus-osd/internal/zfs/zfs.go
```diff
@@ -843,3 +843,52 @@ func ScrubZpool(ctx context.Context, poolName string) error {
...
+					latestPoolInfo = p
+				}
+			}
+
+			// If the scrub is not in progress, break and move to the next pool.
+			if latestPoolInfo.LastScrub != nil && latestPoolInfo.LastScrub.State != api.ScrubInProgress {
+				break
+			}
+
+			time.Sleep(5 * time.Second)
```
gibmat: I'd sleep for at least a minute between checks. Initial empty pools will scrub quickly, but once data starts accumulating a scrub could easily last into the hours and pinging every five seconds seems a bit wasteful.

incus-osd/api/system_storage.go
```diff
@@ -4,7 +4,8 @@ import "time"
 
 // SystemStorageConfig represents additional configuration for the system's local storage.
 type SystemStorageConfig struct {
-	Pools []SystemStoragePool `json:"pools,omitempty" yaml:"pools,omitempty"`
+	ScrubSchedule string              `json:"scrub_schedule"  yaml:"scrub_schedule"`
```
gibmat: Should probably also be marked `omitempty`

gibmat (changes requested):

incus-osd/internal/state/struct.go
stgraber: Oops, I missed that in the review because I saw the `json` tag.
They indeed should have an `incusos:"-"` tag added to them as we don't want (and can't as they're complex structs) persist any of that.

incus-osd/internal/state/struct.go
stgraber: Per my earlier review, I think this can go away completely.

The default value should be handled as a state default and be added during upgrade, at which point we can just rely on the entry in the state struct directly without an access function.

incus-osd/internal/state/struct.go
luissimas: Nice! I'll do that

incus-osd/internal/state/struct.go
luissimas: Makes sense, I'll refactor this. I was actually wondering about what's the preferred way to handle external dependencies in the project.

incus-osd/internal/state/struct.go
luissimas: Oh, nice! I didn't knew about the `incusos` tag.

But I'm a bit confused here. The current encoding logic seems to treat `json` and `incusos` tags equally when omitting fields. So does the `incusos` tag has any particular effect in this case or is it just better for explicitness?

https://github.com/lxc/incus-os/blob/90e92c030b9ea78a8cde572b9fbaa74a68304498/incus-osd/internal/state/encode.go#L112-L114

There are no fields in the `State` struct with the `incusos:"-"` tag, even `UpdateMutex`, which I assume that we also don't want to persist in the state file. In this case, should them also be annotated with `incusos:"-"` ?

https://github.com/lxc/incus-os/blob/90e92c030b9ea78a8cde572b9fbaa74a68304498/incus-osd/internal/state/struct.go#L36-L41

stgraber: @gibmat ^

gibmat: The `incusos` tag was introduced to prevent saving the various state structs in the api definitions to the IncusOS state file but still report those structs via the REST API. (For an example, see https://github.com/lxc/incus-os/blob/main/incus-osd/api/system_network.go#L25)

In this case it is no different than using `json:"-"`, and it would probably be cleaner to continue the use of the `json` tag in the state struct definition.

incus-osd/api/system_storage.go
luissimas: Since we will now initialize this value in the state file and validate it when updating via the API, it should never be empty. Given this, I think it doesn't make sense to add the `omitempty` WDYT?

stgraber: Yeah, we can leave it as it is given it will always have a value, makes `omitempty` a bit pointless if it can never be empty ;)

luissimas:
Thanks for the feedback, folks!

I expect to have this ready for another review tomorrow.

For returning the `scrub_schedule`  config option via the API, is there any preferred method? Currently the endpoint that returns system storage information only fetches data from the ZFS pools, but now we'd have to also fetch a portion from the server state.

stgraber:
> For returning the `scrub_schedule` config option via the API, is there any preferred method? Currently the endpoint that returns system storage information only fetches data from the ZFS pools, but now we'd have to also fetch a portion from the server state.

It should look the same as we have for other endpoints like network.
Basically it should get us a struct with both the `config` and `state`.

The `state` will be as it is today, but now `config` will include the `scrub_schedule` config key.

So something like this:
```
stgraber@castiana:~/Code/lxc/incus-os (git:lxc/main)$ incus admin os system storage show dev-cluster:
WARNING: The IncusOS API and configuration is subject to change

config:
  scrub_schedule: "the value"
state:
  drives:
  - boot: false
    bus: nvme
    capacity_in_bytes: 5.3687107584e+10
    id: /dev/disk/by-id/lvm-pv-uuid-QBz0UL-BnMM-6BI5-rJ6k-kmsd-ddA8-UebU1H
    model_family: ""
    model_name: TrueNAS KVM VM
    remote: true
    removable: false
    serial_number: 09df14fa66ea2b2f5a6e
    smart:
      enabled: true
      passed: true
  - boot: false
    bus: scsi
    capacity_in_bytes: 5.3687107584e+10
    id: /dev/disk/by-id/lvm-pv-uuid-ezwscu-iwPJ-2tMZ-aQ6X-uFtB-xRbl-ndeXhq
    model_family: TrueNAS
    model_name: iSCSI Disk
    remote: false
    removable: false
    serial_number: 81748b33ef5fe02
    smart:
      enabled: false
      passed: true
  - boot: false
    bus: nvme
    capacity_in_bytes: 5.36870912e+10
    id: /dev/disk/by-id/lvm-pv-uuid-o1BaPa-CAVE-UlxL-IufY-Coc5-wxsv-peJjXk
    model_family: ""
    model_name: QEMU NVMe Ctrl
    remote: false
    removable: false
    serial_number: incus_shared
    smart:
      enabled: true
      passed: true
  - boot: false
    bus: nvme
    capacity_in_bytes: 5.36870912e+10
    id: /dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_incus_disk1
    model_family: ""
    model_name: QEMU NVMe Ctrl
    remote: false
    removable: false
    serial_number: incus_disk1
    smart:
      enabled: true
      passed: true
  - boot: true
    bus: nvme
    capacity_in_bytes: 8.589934592e+10
    id: /dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_incus_root
    member_pool: local
    model_family: ""
    model_name: QEMU NVMe Ctrl
    remote: false
    removable: false
    serial_number: incus_root
    smart:
      enabled: true
      passed: true
  pools:
  - devices:
    - /dev/disk/by-id/nvme-QEMU_NVMe_Ctrl_incus_root-part11
    encryption_key_status: available
    last_scrub:
      end_time: "1970-01-01T00:00:00Z"
      errors: 0
      progress: 0.00%
      start_time: "1970-01-01T00:00:00Z"
      state: UNKNOWN
    name: local
    pool_allocated_space_in_bytes: 6.063075328e+09
    raw_pool_size_in_bytes: 4.9928994816e+10
    state: ONLINE
    type: zfs-raid0
    usable_pool_size_in_bytes: 4.9928994816e+10
    volumes:
    - name: incus
      quota_in_bytes: 0
      usage_in_bytes: 2.622210048e+09
      use: incus
    - name: linstor
      quota_in_bytes: 0
      usage_in_bytes: 3.416264704e+09
      use: '-'
```

incus-osd/internal/state/struct.go
luissimas: Made the refactoring and added the new `scheduling` package in 3b790dad6c72dc3b621696ed64a697518d826125. Now the gocron dependency is only referenced within the new package, and it should be significantly easier to add new jobs in the future. If there's anything that's not quite right with the package implementation or its API please let me know!

incus-osd/internal/rest/api_system_storage.go
```diff
@@ -91,9 +92,15 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		// Populate config with the current system config.
+		ret.Config = s.state.System.Storage.Config
```
luissimas, line 96: This was the minimal change way I found to include the current storage config in the API return. I'm not happy with it though. I think `storage.GetStorageInfo` should return `storage.SystemStorageState` instead of `storage.SystemStorage`. We then would combine the state fetch from `storage.GetStorageInfo` with the config from `state.State.System.Storage.Config` to create the final API return.

If this makes sense, I can do this in a follow up PR (as this one is getting a bit bigger than expected). WDYT?

stgraber: Yeah, `GetStorageInfo` should just return the State struct. I wonder if its current behavior comes from us using nested structs initially which would have made returning just the state a bit harder.

stgraber (approved):

stgraber:
Thanks for all the work!

PR #744, opened 2025-12-29
incus-osd/install: Improve system check errors

gibmat:
Improve the error message returned when IncusOS starts but doesn't think it should begin an installation:
* If running from read-only media (CDROM, read-only USB stick), return an error about missing the install seed
* If running from a read-write media that's less than 50GiB in size, return an error
* If neither case is true, return an error that `systemd-repart` has failed

### DIFF
--- a/incus-osd/internal/install/install.go
+++ b/incus-osd/internal/install/install.go
@@ -81,22 +81,29 @@ func CheckSystemRequirements(ctx context.Context, t *tui.TUI) error {
 		}
 	}
 
-	// Check if systemd-repart has failed (we're either running from a read-only or a small USB
-	// stick), or we're running from a CDROM, which normally indicates we're about to start an
-	// install but there's no install seed present.
-	if (systemd.IsFailed(ctx, "systemd-repart") || runningFromCDROM()) && !ShouldPerformInstall() {
-		return errors.New("unable to begin install without seed configuration")
+	// Get the source device that IncusOS is running from.
+	sourceDevice, sourceIsReadonly, sourceDeviceSize, err := getSourceDevice(ctx)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


incus-osd/internal/install/install.go
```diff
@@ -81,22 +81,29 @@ func CheckSystemRequirements(ctx context.Context, t *tui.TUI) error {
...
+		return errors.New("unable to determine source device: " + err.Error())
 	}
 
-	// Perform install-specific checks.
-	if installSeed != nil { //nolint:nestif
-		// Check that we have either been told what target device to use, or that we can automatically figure it out.
-		source, _, err := getSourceDevice(ctx)
-		if err != nil {
-			return errors.New("unable to determine source device: " + err.Error())
+	// If we aren't going to perform an install but systemd-repart failed or we're running from a CDROM,
```
stgraber, line 90: How's that case even possible?

If systemd-repart fails we don't have a root filesystem so we should never even make it into `incus-osd`, or am I missing something here?

gibmat: In the initrd, if systemd-repart fails we automatically mount `/` as a tmpfs. This is needed for running the installer from a CDROM or a small USB drive. Once the install completes on the new system's first boot systemd-repart will then properly setup the remaining partitions and IncusOS will run with a proper `/`.

When running IncusOS live from a large enough disk systemd-repart will be able to create the partitions and we'll start with a proper `/`.

The `CheckSystemRequirements()` function is in the install package; maybe it could get refactored out to somewhere else, but it might be too tightly integrated for that to be an easy task.

stgraber: Oh right, that's how we end up there :)

stgraber (approved):

PR #749, opened 2025-12-31
Support setting custom CA certificates

### LINKED ISSUE #711: Support for custom CA certificates
Can we add support for adding custom CA certificates in to the trust store.

This is for cases where a user is running their own custom public key infrastructure (PKI) or simply have a requirement to trust custom CA certificates.

gibmat:
Closes #711

### DIFF
--- a/incus-osd/cmd/incus-osd/main.go
+++ b/incus-osd/cmd/incus-osd/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/lxc/incus-os/incus-osd/internal/storage"
 	"github.com/lxc/incus-os/incus-osd/internal/systemd"
 	"github.com/lxc/incus-os/incus-osd/internal/tui"
+	"github.com/lxc/incus-os/incus-osd/internal/util"
 	"github.com/lxc/incus-os/incus-osd/internal/zfs"
 )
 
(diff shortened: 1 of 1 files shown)


incus-osd/api/system_security.go
```diff
@@ -13,7 +13,8 @@ type SystemSecurityState struct {
 
 // SystemSecurityConfig holds additional security configuration settings.
 type SystemSecurityConfig struct {
-	EncryptionRecoveryKeys []string `json:"encryption_recovery_keys" yaml:"encryption_recovery_keys"`
+	CustomCAPEMCerts       []string `json:"custom_ca_pem_certs,omitempty" yaml:"custom_ca_pem_certs,omitempty"`
```
stgraber: I don't think we need the format in the field name as we're not likely to support anything but PEM anyway (and if we did, we could easily tell them apart in the list).

Just `CustomCACerts` and `custom_ca_certs` should be easier to remember without having to go look at the struct.

incus-osd/internal/util/ssl.go
```diff
@@ -0,0 +1,86 @@
...
+		if !strings.HasSuffix(pemCert, "\n") {
+			_, err := buf.WriteString("\n")
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	// Write the full set of CA certificates to the file system.
+	return os.WriteFile("/etc/ssl/certs/ca-certificates.crt", buf.Bytes(), 0o644)
```
stgraber: That doesn't look ideal from a memory standpoint as we're dealing with a big text variable and then dumping it back to a file.

Would it make more sense to:
- os.Create the target
- os.Open the source
- io.CopyN (usual loop) from source to target
- WriteString the extra entries

That would avoid ever having to load everything into memory and shouldn't be much more complex as far as logic goes.

doc/rest-api.yaml
```diff
@@ -1,4 +1,8 @@
 definitions:
+    ACMEChallengeType:
```
stgraber, line 2: I usually like having the Swagger YAML be its own commit as it tends to be pretty noisy.

stgraber:
Just a few tiny changes and this will be good to go!

PR #754, opened 2026-01-02
Add system/kernel API endpoint

### LINKED ISSUE #630: Add a way to prepare a PCIE device for passtrought
This one might be more difficult and maybe impossible, as it's change files that need the initramfs to be reloaded (and I don't know if Incus OS allow that)

Like is said in the proxmox [documentation](https://pve.proxmox.com/wiki/PCI(e)_Passthrough#_host_device_passthrough) and something I have already experienced, sometime, PCIE devices are not released from the host to be passtrought which lead to crash or lockup.

With a parameter, like a list of PCIE address for the devices we would want to prepare, Incus OS should : 
- add ` options vfio-pci ids=` with the id to a modprobe conf file
- add `softdep <some-module> pre: vfio-pci` to a modprobe conf file 
- reload initramfs
- tell the admin to reboot the host 

I don't know if it's possible with the architecture of Incus OS but I propose it to see.

### LINKED ISSUE #730: TCP tuning
Hi, this project piqued my interest given its focus on security. My intended use case is to replace Proxmox on an offsite backup server, but I might migrate the main linux server to it as well. 

One concern I have is that the backup server is quite far away (200ms, 300 Mbps), and the default kernel TCP parameters are insufficient to fully utilize the connection due to the window size being capped. Default kernel parameters can only achieve around 60-70 Mbps on a single connection. I have the following settings in my sysctl.conf, which allows the full bandwidth to be used:

```
net.ipv4.tcp_rmem = 4096 65536 33554432
net.ipv4.tcp_wmem = 4096 65536 33554432
net.core.rmem_max = 33554432
net.core.wmem_max = 33554432
net.core.default_qdisc = fq
net.ipv4.tcp_congestion_control = bbr
```

Would it be possible to allow the user to load a sysctl.conf or provide an API to set some of these parameters? I understand that these can be set in VMs or containers, but that does not help with operations like `incus copy`.

gibmat:
Add a `system/kernel` API endpoint that allows setting pass-through PCI(e) device IDs, blacklisting kernel modules, and tweaking some select `sysctl` network values for an IncusOS system.

Closes #630, Closes #730

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -64,6 +64,7 @@ OVMF
 OVN
 OVS
 parsable
+PCI
 PCR
 PCRs
 PEM
--- a/doc/reference/system.md
+++ b/doc/reference/system.md
@@ -11,6 +11,7 @@ as well as relevant actions.
 :maxdepth: 1
 
 Backup/Restore </reference/system/backup>
+Kernel </reference/system/kernel>
 Logging </reference/system/logging>
 Network </reference/system/network>
 Power </reference/system/power>
--- a/doc/reference/system/kernel.md
+++ b/doc/reference/system/kernel.md
@@ -1,0 +1,24 @@
+# Kernel
+
+IncusOS exposes a limited set of configuration knobs for adjusting kernel-level settings. A system reboot may be required for changes to fully take effect.
+
+## Configuration options
+
+Configuration fields are defined in the [`SystemKernelConfig` struct](https://github.com/lxc/incus-os/blob/main/incus-osd/api/system_kernel.go).
+
+The following configuration options can be set:
+
+* `blacklist_modules`: A list of one or more kernel modules to blacklist. Typically useful when passing through PCI devices to virtual machines.
+
+* `network`: Change `sysctl` values that impact the system's network configuration.
+   * `buffer_size`: Optional; configure the maximum buffer size used when setting the `net.ipv4.tcp_rmem`, `net.ipv4.tcp_wmem`, `net.core.rmem_max`, and `net.core.wmem_max` `sysctl` fields.
+
+   * `queuing_discipline`: Optional; configure the value of the `net.core.default_qdisc` `sysctl` field.
+
+   * `tcp_congestion_algorithm`: Optional; configure the TCP congestion algorithm used by the system, defaults to `bbr`.
+
+* `pci`: Change PCI device configuration.
+   * `passthrough`: Configure one or more PCI devices for pass-through to a virtual machine:
+      * `vendor_id`: The PCI vendor ID
+      * `product_id`: The PCI product ID
+      * `pci_address`: Optional; if specified the system will attempt to unbind the given PCI device from its existing driver and configure it for passing though to a virtual machine without requiring a reboot.


incus-osd/internal/kernel/kernel.go
```diff
@@ -0,0 +1,163 @@
...
+	"strings"
+
+	"github.com/lxc/incus-os/incus-osd/api"
+	"github.com/lxc/incus-os/incus-osd/internal/systemd"
+)
+
+// ApplyKernelConfiguration updates various parts of the kernel configuration. A reboot
+// may be required to fully apply the changes.
+func ApplyKernelConfiguration(ctx context.Context, config api.SystemKernelConfig) error {
+	// Update the list pf PCI(e) pass-throughs.
```
stgraber: Typo here s/pf/of/

incus-osd/internal/kernel/kernel.go
```diff
@@ -0,0 +1,163 @@
...
+
+func updateBlacklistModules(modules []string) error {
+	// Remove the existing configuration file, if it exists.
+	_, err := os.Stat("/etc/modprobe.d/99-blacklist-modules.conf")
+	if err == nil {
+		err = os.Remove("/etc/modprobe.d/99-blacklist-modules.conf")
+		if err != nil {
+			return err
+		}
+	}
```
stgraber: It'd be a cheaper call to just hit `os.Remove` and handle the NotExist error (saves one potential syscall in the case where it exists).

incus-osd/internal/kernel/kernel.go
```diff
@@ -0,0 +1,163 @@
...
+		return err
+	}
+
+	// Write the file contents.
+	_, err = fd.WriteString("options vfio-pci ids=" + strings.Join(pciIDs, ",") + "\n")
+
+	return err
+}
+
+func updateBlacklistModules(modules []string) error {
```
stgraber, line 44: Should we try to unload any listed kernel module in there?
I don't think we should fail on failure, but that may save a reboot for some folks.

incus-osd/internal/kernel/kernel.go
```diff
@@ -0,0 +1,163 @@
...
+	// Update local sysctl configuration.
+	err = updateSysctlConfig(ctx, config.SysctlConfig)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func updatePCIPassthroughs(pciIDs []string) error {
```
stgraber: I wonder if it'd make sense to attempt to unbind+bind matching devices so that we can save ourselves the reboot in most cases.

stgraber:
On top of the in-line comments, I have some concerns about those knobs being able to seriously break the system to a point that can't be easily recovered from.

Do you think there's a reasonable way for us to disable all of those options when booting from the recovery (older) system image?

stgraber:
On the sysctl feature, I'm pretty conflicted by it...

On the one hand, it definitely solves a current problem by making it possible to address less common configuration issues like the one in #730. On the other hand, I really would prefer IncusOS be able to just do the right thing out of the box in the vast majority of cases and when that's not possible, offer knobs that make sense to users rather than exposing the entire set of kernel settings.

I also have a bit of a concern about a number of sysctls making the machine entirely unreachable if not unbootable and some of them allowing for some pretty bad stuff which would bypass our normal security measures :(

I don't know if we should stay away from it and just turn it into a list of low level settings that we support, effectively providing a thin abstraction over them? That'd require folks file issues for anything else they're missing, but on the upside, that would result in a better user experience for everyone after that.


Looking back at #730, I think I'd:
 - Make `bbr` the default TCP congestion algorithm (that's what I've done on my systems and it does result in better performance over high latency with no visible impact on low latency)
 - Have an option to control the TCP congestion algorithm
 - Have an option to control the qdisc
 - Have an option to control the network buffer size (bumps all 4 rmem/wmem keys)

So maybe something like:
```go
type Settings struct {
    Network struct {
        TCPCongestionAlgorithm string
        QueuingDiscipline string
        BufferSize int
    }
}
```

That would give us a reasonable framework to add more kernel options while not exposing the entire set of sysctl and letting us do a bit of validation on the configuration.

@gibmat what do you think?

incus-osd/internal/kernel/kernel.go
gibmat: Yeah, easy enough to try an `rmmod` for each blacklisted module.

incus-osd/internal/kernel/kernel.go
gibmat: Based on some Google results, I think I've got the logic to opportunistically attempt to unbind/bind the PCI device, but we should probably test on real hardware to make sure it works properly.

gibmat:
> I don't know if we should stay away from it and just turn it into a list of low level settings that we support, effectively providing a thin abstraction over them? That'd require folks file issues for anything else they're missing, but on the upside, that would result in a better user experience for everyone after that.

I've removed the ability to directly set `sysctl` values based on our conversation. I changed the API to have some additional configuration fields for network and PCI, from which we can then generate the appropriate configuration.

stgraber:
@gibmat I think we're just missing a reference doc page for this.

PR #756, opened 2026-01-03
image-customizer: Get exact username from Discourse

stgraber:


stgraber:
We've seen a couple of users with mismatching case running into issues.

PR #758, opened 2026-01-04
incus-osd: Refactor GetStorageInfo to return only the state

luissimas:
# Description

As discussed in [this comment](https://github.com/lxc/incus-os/pull/739#discussion_r2659104977), the `storage.GetStorageInfo` should return `storage.SystemStorageState` instead of `storage.SystemStorage`.

I've done the changes in a single commit to keep the refactoring atomic. If it is required, I can split the changes in a set of commits per package while making sure all of them compile.

This should be a pure refactor: no changes in external behavior.

### DIFF
--- a/incus-osd/internal/rest/api_system_storage.go
+++ b/incus-osd/internal/rest/api_system_storage.go
@@ -85,15 +85,17 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 
 	switch r.Method {
 	case http.MethodGet:
-		ret, err := storage.GetStorageInfo(r.Context())
+		info, err := storage.GetStorageInfo(r.Context())
 		if err != nil {
 			_ = response.InternalError(err).Render(w)
 
 			return
 		}
 
-		// Populate config with the current system config.
-		ret.Config = s.state.System.Storage.Config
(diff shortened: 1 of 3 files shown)


incus-osd/internal/rest/api_system_storage.go
```diff
@@ -140,13 +134,7 @@ func (s *Server) apiSystemStorage(w http.ResponseWriter, r *http.Request) {
 
 		// Create or update a pool.
 		if len(storageStruct.Config.Pools) == 0 {
-			if len(current.Config.Pools) == 0 {
-				_ = response.EmptySyncResponse.Render(w)
-
-				return
-			}
```
luissimas, line 147: This condition was always evaluated to true, as `current.Config.Pools` was always empty due to `current.Config` never being initialized in `storage.GetStorageInfo`.

If I understood correctly, the idea is that a PUT sent with an empty list of pools should return a success so we don't force the user to specify the entire list of pools for every change. If this is the case, then this behavior is kept by this change.

stgraber: Yeah, looks fine

stgraber (approved):

PR #776, opened 2026-01-07
Implement internal authentication token API

stgraber:


incus-osd/internal/rest/server.go
```diff
@@ -49,6 +49,7 @@ func (s *Server) Serve(ctx context.Context) error {
 	router := http.NewServeMux()
 
 	router.HandleFunc("/", s.apiRoot)
+	router.HandleFunc("/internal/tui/:write-message", s.apiInternalTUI)
```
gibmat, line 53: We'll also need to update `incusos_tests/tests_incusos_api_debug.py`.

incus-osd/internal/rest/api_internal.go
```diff
@@ -13,6 +13,12 @@ import (
 func (*Server) apiInternalTUI(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
+	if r.Header.Get("X-IncusOS-Proxy") != "" {
```
gibmat, line 17: As long as we're only relying on this to stop accidental use of the internal API, I think it's OK. But we shouldn't rely on it to actually protect any API from a malicious user.

incus-osd/internal/auth/token.go
```diff
@@ -0,0 +1,186 @@
...
+
+const (
+	ctxPrimaryPath = "/run/incus-os/tpm-auth.primary.ctx"
+	ctxKeyPath     = "/run/incus-os/tpm-auth.key.ctx"
+
+	privatePath = "/var/lib/incus-os/tpm-auth.priv"
+	publicPath  = "/var/lib/incus-os/tpm-auth.pub"
+	pemPath     = "/var/lib/incus-os/tpm-auth.pem"
+
+	authCertPath = "/usr/lib/incus-osd/certs/auth.crt"
```
gibmat, line 26: I'd personally use a more descriptive name than just `auth.crt` for this file; something like `token-encryption.crt` or similar.

incus-osd/internal/auth/token.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	machineIDStr := strings.TrimSpace(string(machineID))
+
+	_, _ = out.WriteString(machineIDStr + ":")
+
+	// Prepare the token.
+	type token struct {
```
gibmat: I expect we'll be consuming this struct in other Go code, so it would probably be best for it to be a properly exported type from the package so we don't have two (or more) versions floating around.

incus-osd/internal/auth/token.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Get the system machine ID.
+	machineID, err := os.ReadFile("/etc/machine-id")
+	if err != nil {
+		return "", err
+	}
+
+	machineIDStr := strings.TrimSpace(string(machineID))
+
+	_, _ = out.WriteString(machineIDStr + ":")
```
gibmat: Minor nit, here and down below when flushing writes we don't really expect to encounter an error, but since this is part of what I'd consider security-sensitive code, we probably shouldn't ignore errors if they happen.

incus-osd/internal/auth/token.go
stgraber: I expect that to go away in a few days as we switch to the JSON keys stuff :)
We won't actually be shipping any build with an `auth.crt` until then as that's basically just going to be used for HypervisorOS.

incus-osd/internal/rest/api_internal.go
stgraber: It should be good enough. Whenever you're going through Incus, Operations Center or Migration Manager, the header will be set.

If you're not, then you somehow have found a way to get access to the unix socket directly which is only accessible by root on the host, so we'd have bigger problems at this point ;)

But yeah, the two endpoints we currently have are both technically find for anyone to access, they just don't make much sense outside of the local use, so the restriction is there to keep folks away from poking at /internal/XYZ remotely.

stgraber:
@gibmat I think that addressed the feedback so far

gibmat (approved):

Issue #743, opened 2025-12-29
Introduce a `keys.json` or similar file containing all the keys we use


stgraber:
`incus-osd` currently needs quite a few public keys to function properly:
- SecureBoot kernel keyring identifiers (2025 & 2026)
- Update signing sub-CA (update and hotfix validation)

In addition to that, @gibmat was saying we'll also need a full version of both of the SecureBoot keys and some work I'm planning is going to need us to handle an additional couple of keys:
- Support for a different sub-CA for signing of hotfix/debug scripts
- Support for an authentication key to which we'd encrypt some authentication data

Rather than keep spreading that across the `incus-osd` codebase, having a single `keys.json` file in the codebase containing all the relevant keys would make sense. We'd have the repository ship with the production keys but that'd allow for easily overriding it with local test values for manual builds as well as allow for downstreams like FuturFusion's HypervisorOS to easily swap the key configuration without needing code changes.

PR #780, opened 2026-01-08
Update recovery logic and add tests

gibmat:
* Refactor/fix issues found while getting a locally-built IncusOS running on my physical system.
* Properly update application versions.
* Don't allow any version older than the currently running system to install.
* Add a `GetSigningCACert()` method to each provider which will return the appropriate CA used to verify update signatures. Once #743 is completed, we can probably have a single common method that retrieves the correct cert.
* Add tests exercising the ability to apply updates via the recovery mechanism.


incus-osd/internal/providers/provider_operations_center.go
```diff
@@ -183,6 +183,10 @@ func (*operationsCenter) Type() string {
 	return "operations-center"
 }
 
+func (*operationsCenter) GetSigningCACert() (string, error) {
```
stgraber, line 186: @gibmat we'd want that to return the same as from the images provider.

Otherwise we'll never be able to use the recovery feature on a system that's managed by operations center.

stgraber:
@gibmat needs a rebase

PR #782, opened 2026-01-09
base: Add more MMC modules

stgraber:


stgraber:
Landing that so we have it in the build I'm about to tag.
It's not like the tests do anything about the initrd content anyway.

PR #783, opened 2026-01-09
Fix app-build version handling

stgraber:


incus-osd/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/FuturFusion/operations-center v0.2.0
```
gibmat: Missing a version bump to 0.2.1.

stgraber: Oh, weird that it didn't pick it up, that's why I ran the update-gomod in the first place for...

stgraber:
@gibmat also added a gpu-support start function to tidy things up there

gibmat (approved):

stgraber:
@gibmat one more fix added to the pile

PR #791, opened 2026-01-11
incus-osd/systemd/networkd: Fix swapped GRO, GSO

jasisk:
[Context](https://discuss.linuxcontainers.org/t/nic-no-longer-responsive-to-wake-on-lan-when-powering-off-from-incusos/25849/14?u=grip).

I think these configs are swapped.

(Not an LLM but forgot I had copilot auto review on. Sorry about that!)


stgraber:
Looks good, thanks!

stgraber:
Looks like the fields show up in a slightly different order so the test needs adjusting.

jasisk:
> Looks like the fields show up in a slightly different order so the test needs adjusting.

Fixed. CI as test runner since I'm flying blind here. 😀

stgraber:
Thanks!

PR #793, opened 2026-01-11
Implement image server authentication

stgraber:
This adds a few things:
 - It adds a new token config option to the images provider
 - It adds support for registration to the images provider when a token is present in config
 - It switches the default HTTP client for one setting the required authentication header when a token is present in config
 - It cleans up our use of `/etc/machine-id` in the codebase for a shared implementation through the state struct
 - It adds an additional internal function to fetch a registration token struct

For context, we don't expect to use any of this with the public image server. All of this is designed for use by HypervisorOS with FuturFusion's private image server, but since the concept is pretty generic, we're landing that in IncusOS so others may use it for their own private builds and image servers.

In our case, we need to handle two cases:
 - Automated registration for systems with Internet access. Those will hit the /register endpoint on first boot, at which point the `X-IncusOS-Authentication` header is used to authenticate the system with any subsequent query to the image server. The `generate-token` internal API endpoint can then be used by Operations Center to similarly authenticate with the image server without it having to directly mess with the TPM.
 - Offline registration for systems without Internet access. Those systems will boot with the local provider and install Operations Center that way. The resulting Operations Center won't have any way to fetch updates, but it will be able to hit `generate-registration` to provide a valid registration JSON struct that the user can directly provide to the image server through a separate online system. Once the registration is done, manual image downloads can be performed by getting an authentication token from Operations Center and once again providing that to the image server through a separate online system. The downloaded update files can then be loaded into the offline Operations Center.

We use a TPM signed token including a timestamp and machine ID as a way to reliably authenticate the system even in offline mode with something that can't be spoofed by another machine and which should allow for detecting cloned systems even if they rely on a software TPM.

### DIFF
--- a/incus-osd/internal/rest/api_internal.go
+++ b/incus-osd/internal/rest/api_internal.go
@@ -55,6 +55,57 @@ func (*Server) apiInternalTUI(w http.ResponseWriter, r *http.Request) {
 	_ = response.EmptySyncResponse.Render(w)
 }
 
+func (s *Server) apiInternalRegistration(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	if r.Header.Get("X-IncusOS-Proxy") != "" {
+		_ = response.Forbidden(nil).Render(w)
+
+		return
+	}
+
+	if r.Method != http.MethodPost {
+		_ = response.NotImplemented(nil).Render(w)
+
+		return
+	}
+
+	type apiPost struct {
+		Token string `json:"token"`
+	}
+
+	req := &apiPost{}
+
+	err := json.NewDecoder(r.Body).Decode(req)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


incus-osd/internal/auth/token.go
```diff
@@ -13,6 +13,8 @@ import (
 	"time"
 
 	"github.com/lxc/incus/v6/shared/subprocess"
+
+	apiupdate "github.com/lxc/incus-os/incus-osd/api/images"
```
gibmat, line 17: Wouldn't `apiimages` be less confusing?

gibmat (approved):

incus-osd/internal/auth/token.go
stgraber: It would if we hadn't been using `apiupdate` for that everywhere else in the codebase :)

PR #794, opened 2026-01-11
Restore MAC addresses for WoL enabled NICs on shutdown

stgraber:


incus-osd/cmd/incus-osd/main.go
```diff
@@ -604,6 +604,7 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error { //nolint:r
 		case "shutdown":
 			_ = systemd.SystemPowerOff(ctx)
 		case "reboot":
+			systemd.RestoreWOLMACAddresses(ctx, s)
```
jasisk: Should this be for `shutdown`?

stgraber: Oops, fixed!

incus-osd/internal/systemd/networkd.go
```diff
@@ -392,6 +392,59 @@ func UpdateNetworkState(ctx context.Context, n *api.SystemNetwork) error {
...
+	}
+
+	for _, i := range s.System.Network.Config.Interfaces {
+		if !i.Ethernet.WakeOnLAN {
+			continue
+		}
+
+		iface := "_p" + strings.ToLower(strings.ReplaceAll(i.Hwaddr, ":", ""))
+
+		err := restoreMAC(iface)
```
gibmat: Couldn't we just do a `restoreMAC(iface, i.Hwaddr)` and skip the extra `ip` command getting the permanent MAC for the physical device, which we already know?

incus-osd/internal/systemd/networkd.go
```diff
@@ -2028,21 +2081,26 @@ func resolveMACs(ctx context.Context, config *api.SystemNetworkConfig) error {
...
-		return "", err
+		return "", "", err
 	}
 
-	match := macAddressRegex.FindAllStringSubmatch(output, -1)
-	if len(match) != 1 {
-		return "", errors.New("no MAC address found")
+	link := &linkJSON{}
+
+	err = json.Unmarshal([]byte(output), link)
```
gibmat: This won't work without a `-j` given to the `ip` command, and its json output is an array of objects even when a single interface is specified.

gibmat (changes requested):

incus-osd/internal/systemd/networkd.go
stgraber: Oops, yeah, you're right, it's a bunch of extra work for nothing ;)

PR #797, opened 2026-01-13
incus-osd/applications/gpu-support: Unbind the devices

stgraber:
This is needed at least for i915.

### DIFF
--- a/incus-osd/internal/applications/app_gpu_support.go
+++ b/incus-osd/internal/applications/app_gpu_support.go
@@ -2,8 +2,11 @@ package applications
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/lxc/incus/v6/shared/subprocess"
 )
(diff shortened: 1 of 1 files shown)


incus-osd/internal/applications/app_gpu_support.go
```diff
@@ -25,6 +28,29 @@ func (*gpuSupport) Start(ctx context.Context, _ string) error {
...
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			slog.Warn("Failed to check kernel module", "module", module)
+
+			continue
+		}
+
+		for _, entry := range entries {
+			address := filepath.Base(entry.Name())
+
+			if !strings.Contains(address, "") || !strings.Contains(address, ".") {
```
gibmat: Won't `!strings.Contains(address, "")` always be false?

stgraber: oh, oops, not sure what happen to this, it should have been `":"`, let me fix.

stgraber: Fixed!

PR #827, opened 2026-01-20
Add support for EDK2 logo replacement

stgraber:
This is logic we're adding to IncusOS but which won't actually be used in production IncusOS builds. Instead this will be used in downstream projects like HypervisorOS where replacing the logo makes sense from a branding standpoint.


mihalicyn (approved):
Cool stuff!

PR #828, opened 2026-01-20
incus-osd/recovery: Fix installation of applications

stgraber:
This improves the handling of the recovery disk as a way to install applications when no internet connectivity is present.

It fixes getting the initial application list by respecting the seed data and it skips over OS and application files which are already locally present allowing for a much smaller rescue disk to be used.

### DIFF
--- a/incus-osd/internal/recovery/recovery.go
+++ b/incus-osd/internal/recovery/recovery.go
@@ -22,6 +22,7 @@ import (
 	"github.com/lxc/incus-os/incus-osd/api"
 	apiupdate "github.com/lxc/incus-os/incus-osd/api/images"
 	"github.com/lxc/incus-os/incus-osd/internal/providers"
+	"github.com/lxc/incus-os/incus-osd/internal/seed"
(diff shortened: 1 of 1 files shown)


incus-osd/internal/recovery/recovery.go
```diff
@@ -249,9 +266,22 @@ func applyUpdate(ctx context.Context, s *state.State, updateCA string, mountDir
...
 		if file.Type == apiupdate.UpdateFileTypeApplication {
-			if !slices.Contains(installedApplications, filepath.Base(strings.TrimSuffix(file.Filename, ".raw.gz"))) {
+			appName := filepath.Base(strings.TrimSuffix(file.Filename, ".raw.gz"))
+
+			// Don't process any applications that are not meant to be installed.
+			if !slices.Contains(installedApplications, appName) {
+				continue
+			}
+
+			// Don't both with an app update if already on the same version.
```
gibmat: Minor typo in comment (both -> bother)

PR #831, opened 2026-01-20
incus-osd/tests: Run tests in parallel

gibmat:
Now that we've got faster hardware, let's try running the daily tests in parallel so the job doesn't take so long to finish.


stgraber:
@gibmat we should bump the resources for the job too.
We currently just have 4 vCPUs and 8GB of RAM I believe.

We should bump to 8 vCPUs and 32GB of RAM to be safe.

PR #836, opened 2026-01-21
Update handling of `update` seed

gibmat:
* Move validation for the `update` seed to a common function
* Handle the case were we might be provided an empty `update` seed and copy over the `Channel` and `CheckFrequency` from the existing configuration, if provided.

### DIFF
--- a/incus-osd/internal/seed/update.go
+++ b/incus-osd/internal/seed/update.go
@@ -3,17 +3,32 @@ package seed
 import (
 	"context"
 
+	"github.com/lxc/incus-os/incus-osd/api"
 	apiseed "github.com/lxc/incus-os/incus-osd/api/seed"
 )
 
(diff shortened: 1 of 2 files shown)


incus-osd/internal/seed/update.go
```diff
@@ -16,6 +17,20 @@ func GetUpdate(_ context.Context) (*apiseed.Update, error) {
 		return nil, err
 	}
 
+	// If an existing configuration was provided and the seed is missing either
+	// the Channel or CheckFrequency fields, copy the existing values. We can't
+	// differentiate an update or missing values for AutoReboot and
+	// MaintenanceWindows, so those will always overwrite any previously set value(s).
```
stgraber, line 23: It doesn't matter as those can't have been set prior to us running GetUpdate anyways. The only two keys that have a value through the state default logic are Channel and CheckFrequency.

PR #850, opened 2026-01-23
Tweak to BMC virtual disk filtering logic

stgraber:


stgraber:
@gibmat this will make the `incus admin os system storage show` output a bit cleaner on my servers ;)

PR #856, opened 2026-01-27
Limit logs written to `/var/log/`

### LINKED ISSUE #704: Add support for log rotation (or force through journal)
Currently we have a bunch of stuff logging to /var/log/ but with no rotation policies in place which will eventually lead to us running out of space:

```
/var/log/sanlock.log
/var/log/incus/dnsmasq.meshbr0.log
/var/log/incus/v3/qemu.qmp.log
/var/log/incus/dnsmasq.incusbr0.log
/var/log/incus/incusd.log
/var/log/incus/internal_ovn-central/lxc.log
/var/log/linstor-satellite/ErrorReport-691D822C-CAFFD-000000.log
/var/log/linstor-satellite/ErrorReport-691D822C-CAFFD-000002.log
/var/log/linstor-satellite/error-report.mv.db
/var/log/linstor-satellite/ErrorReport-691D822C-CAFFD-000001.log
/var/log/openvswitch/ovs-vswitchd.log
/var/log/openvswitch/ovsdb-server.log
```

We should go over every application and service and make sure that:

 - Whenever possible, they exclusively log through the systemd journal
 - For the remaining cases, we have either:
   - The application/service already rotating the files
   - We have log rotation configured externally (likely through `logrotate`)

So far from what I'm aware:

- [ ] `incus` should be configured to only log through journald (stdout/stderr)
- [ ] `migration-manager` should be configured to only log through journald (stdout/stderr)
- [ ] `operations-center` should be configured to only log through journald (stdout/stderr)
- [ ] We need to confirm that Incus clears/rotates the `dnsmasq` log files on daemon restart
- [ ] `openvswitch` should be configured to only log through journald (stdout/stderr)
- [ ] We need to figure out how Linstor logging/error reports work and see if we can send everything through journald
- [ ] `sanlock` should be configured to only log through journald (stdout/stderr)

TBD:

- `iscsi`
- `multipath`
- `tailscale`

gibmat:
Changes:
* Prevent the incus, migration-manager, and operations-center applications from writing logs to `/var/log/` and rely on capturing stdout/stderr from each service to the journal
  * incus does truncate its dnsmasq log(s) on each start
* Prevent OVN from writing to log files and capture stdout/stderr to the journal; because of how the service is defined those files will still be created, but will be empty
* Prevent sanlock from writing to a log file and capture stdout/stderr to the journal; because of how the daemon operates an empty log file will still be created but won't be written to
* Cause linstor-satellite to write its error report dumps to a directory under `/tmp/`. There's no way to disable creation of these error reports, and I don't know if it's worthwhile to setup a log rotation policy just for them. The service itself is logging to the journal.

I also looked at the iscsi and multipath services, which I can start but can't really test. By default they don't seem to write logs to `/var/log/`.

I partially tested tailscale, but since I don't have a test account couldn't fully test to see if any extra log files are created. But default the service logs to the journal.

Closes #704

### DIFF
--- a/incus-osd/internal/services/service_ovn.go
+++ b/incus-osd/internal/services/service_ovn.go
@@ -105,8 +105,16 @@ func (n *OVN) Start(ctx context.Context) error {
 		return nil
 	}
 
+	// Disable writing log output to files, and bump syslog level. Because of how the systemd services
+	// are defined, empty log files will be created under /var/log/openvswitch/.
+	err := os.WriteFile("/etc/default/openvswitch-switch", []byte(`OVS_CTL_OPTS='--ovsdb-server-options="-vsyslog:info -vfile:off" --ovs-vswitchd-options="-vsyslog:info -vfile:off"'
+`), 0o644)
+	if err != nil {
+		return err
+	}
+
 	// Start OVS.
-	err := systemd.StartUnit(ctx, "ovs-vswitchd.service")
+	err = systemd.StartUnit(ctx, "ovs-vswitchd.service")
 	if err != nil {
 		return err
 	}


incus-osd/internal/services/service_ovn.go
```diff
@@ -65,18 +65,17 @@ func (n *OVN) Update(ctx context.Context, req any) error {
...
-	// Enable the service if requested.
+	// Enable and configure the service if requested.
 	if !oldState.Config.Enabled && newState.Config.Enabled {
 		err := n.Start(ctx)
 		if err != nil {
 			return err
 		}
-	}
 
-	// Configure the service.
```
stgraber, line 76: Hmm, that looks wrong. What happens if I'm not enabling or disabling the service, just re-configuring the running service?

With your change, I believe it would turn into a no-op with configure() no longer getting called.

gibmat: Good catch, updated to only configure the service if the new state is enabled.

PR #858, opened 2026-01-28
Collect and embed various certificates used by IncusOS

### LINKED ISSUE #743: Introduce a `keys.json` or similar file containing all the keys we use

gibmat:
Closes #743

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,16 +58,10 @@ jobs:
 
       - name: Load the signing keys
         run: |
-          echo "${{ secrets.SB_CERT }}" > mkosi.crt
-          chmod 644 mkosi.crt
+          ln -s incus-osd/certs/files/secureboot-DB-2026-R1.crt mkosi.crt
 
           echo "${{ secrets.SB_KEY }}" > mkosi.key
           chmod 600 mkosi.key
-
-          mkdir -p certs
-          cd certs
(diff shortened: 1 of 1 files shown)


incus-osd/certs/files/secureboot-DB-2025-R1.crt
```diff
@@ -0,0 +1,16 @@
+-----BEGIN CERTIFICATE-----
```
gibmat, line 1: The db certificate files have been updated to include `-DB` in their name to make things consistent with the PK and KEK files.

PR #873, opened 2026-02-04
Allow arbitrary channels

stgraber:


incus-osd/api/system_update.go
```diff
@@ -41,11 +40,6 @@ type SystemUpdateMaintenanceWindow struct {
 
 // Validate performs basic sanity checks against update configuration.
 func (c *SystemUpdateConfig) Validate() error {
-	// Check the update channel is valid.
```
breml, line 44: We might want to at least check, that it is not empty.

stgraber: Empty strings are actually supported by the code, it will pick the newest image out of any channel it can see.

PR #888, opened 2026-02-10
incus-osd: Tweak GetPrimary logic

stgraber:
This allows selectively requiring an initialized primary application.

### DIFF
--- a/incus-osd/internal/applications/load.go
+++ b/incus-osd/internal/applications/load.go
@@ -36,11 +36,11 @@ func Load(_ context.Context, s *state.State, name string) (Application, error) {
 	return app, nil
 }
 
-// GetPrimary returns the current primary application once initialized.
-func GetPrimary(ctx context.Context, s *state.State) (Application, error) {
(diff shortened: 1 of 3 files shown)


incus-osd/internal/applications/load.go
```diff
@@ -36,11 +36,11 @@ func Load(_ context.Context, s *state.State, name string) (Application, error) {
 	return app, nil
 }
 
-// GetPrimary returns the current primary application once initialized.
-func GetPrimary(ctx context.Context, s *state.State) (Application, error) {
+// GetPrimary returns the current primary application (optionally checking if initialized).
+func GetPrimary(ctx context.Context, s *state.State, initialized bool) (Application, error) {
```
gibmat: I'd suggest renaming that parameter to `requireInitialized`, since it makes things a bit easier to understand.

gibmat (approved):

incus-osd/internal/applications/load.go
stgraber: Done!

PR #894, opened 2026-02-13
Implement drive encryption

### LINKED ISSUE #865: Add simple support for LUKS-encrypted drives to storage APIs
Add basic importing, creation and deletion of LUKS-encrypted drives to the storage API. This will allow IncusOS to manage and automatically unlock encrypted block devices similar to how it does for zpools.

If a drive is encrypted via LUKS, we should report back the device mapper path in the storage info struct.

IncusOS won't do anything additional after unlocking the drive -- it will be up to the container/VM that the drive is attached to further configure the raw block device.

stgraber:
Closes #865

### DIFF
--- a/doc/reference/system/storage.md
+++ b/doc/reference/system/storage.md
@@ -146,3 +146,28 @@ incus admin os system storage delete-volume -d '{"pool":"local","name":"my-volum
 ```{note}
 IncusOS automatically creates a new `incus` volume when setting up the `local` storage pool.
 ```
+
+## Raw drive encryption
+
+IncusOS supports encrypting disks that aren't part of one of its pools.
+This may be useful when providing a container or virtual-machine with full block devices.
+
+IncusOS will generate a random key, use it to encrypt the drive using LUKS and then automatically unlocking the drive on boot.
+Existing encrypted drives can also be added to the system by providing IncusOS with their encryption key.
+
+Wipe and encrypt drive `scsi-0QEMU_QEMU_HARDDISK_incus_disk` by running:
+
+```
+./incus admin os system storage encrypt-drive -d '{"id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk"}'
+```
+
+Import and decrypt an existing drive by running:
+
+```
+./incus admin os system storage import-encrypted-drive -d '{"id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk", "key": "BASE64-OF-KEYFILE"}'
+```
+
+Encrypted drives that are managed by IncusOS get two additional properties in `incus admin os system storage show`:
+
+* Encrypted (indicates whether the drive is encrypted)
+* Encrypted ID (provides the path to the decrypted block device)


incus-osd/internal/storage/luks.go
```diff
@@ -0,0 +1,186 @@
...
+	devName := filepath.Base(devPath)
+	keyfilePath := "/var/lib/incus-os/luks." + devName + ".key"
+
+	// Wipe the drive.
+	err := WipeDrive(ctx, devPath)
+	if err != nil {
+		return err
+	}
+
+	// Generate a random encryption key.
```
gibmat, line 33: It would probably be a good idea to create a helper function that creates a random key given a filepath, so we don't have direct copy-n-paste code here and in the zfs package. :)

incus-osd/api/system_storage.go
```diff
@@ -112,6 +112,17 @@ type SystemStorageWipe struct {
 	ID string `json:"id" yaml:"id"`
 }
 
+// SystemStorageDecrypt defines a struct with information about what drive to decrypt.
```
gibmat: I'd re-order the commits so this one comes before the preceding one, otherwise a git-bisect could end in an un-buildable state.

incus-osd/internal/rest/api_system_storage.go
```diff
@@ -630,3 +630,131 @@ func (*Server) apiSystemStorageScrubPool(w http.ResponseWriter, r *http.Request)
...
+//	      type: object
+//	      example: {"id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk", "key": "KEY-VALUE"}
+//	responses:
+//	  "200":
+//	    $ref: "#/responses/EmptySyncResponse"
+//	  "400":
+//	    $ref: "#/responses/BadRequest"
+//	  "500":
+//	    $ref: "#/responses/InternalServerError"
+func (*Server) apiSystemStorageDecryptDrive(w http.ResponseWriter, r *http.Request) {
```
gibmat: I'd say that this is really more of an "import existing LUKS drive" rather than a "decrypt drive" endpoint. It think it would be a lot less confusing down the road to try to mimic what we've done for the `/1.0/system/storage/:import-pool` endpoint here.

incus-osd/cmd/incus-osd/main.go
```diff
@@ -535,6 +535,11 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error { //nolint:r
 	// Ensure any locally-defined pools are available.
 	slog.InfoContext(ctx, "Bringing up the local storage")
 
+	err = storage.DecryptDrives(ctx)
```
gibmat: Now that we've got two types of local storage, maybe break it out to a separate helper function, just to help logically contain any local storage code?

incus-osd/internal/storage/storage.go
```diff
@@ -718,6 +733,8 @@ func GetStorageInfo(ctx context.Context) (api.SystemStorageState, error) {
 			WWN:             wwnString,
 			SMART:           smartStatus,
 			MemberPool:      driveZpool,
+			Encrypted:       encrypted,
```
gibmat, line 736: Is there a way to either unify/populate these two fields for ZFS drives as well, or rename the fields to be LUKS-specific?

gibmat (changes requested):

incus-osd/internal/storage/storage.go
stgraber: Those fields only make sense when the individual drive/block is encrypted.

For ZFS, the pool is what's encrypted, so if we were to expose Encrypted=true it would be at the pool level for those.

incus-osd/internal/rest/api_system_storage.go
stgraber: import-encrypted-drive maybe?

stgraber:
Alright, that should be the review feedback addressed.

stgraber:
@gibmat anything else needed on this one?

doc/reference/system/storage.md
```diff
@@ -146,3 +146,28 @@ incus admin os system storage delete-volume -d '{"pool":"local","name":"my-volum
...
+Wipe and encrypt drive `scsi-0QEMU_QEMU_HARDDISK_incus_disk` by running:
+
+```
+./incus admin os system storage encrypt-drive -d '{"id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk"}'
+```
+
+Import and decrypt an existing drive by running:
+
+```
+./incus admin os system storage encrypt-drive -d '{"id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_disk", "key": "BASE64-OF-KEYFILE"}'
```
gibmat: Should be `import-encrypted-drive`?

stgraber: oops, yeah, let me fix

stgraber: Fixed!

PR #897, opened 2026-02-16
Add support for system suspend

stgraber:


incus-osd/internal/systemd/system.go
```diff
@@ -25,3 +25,13 @@ func SystemReboot(ctx context.Context) error {
 
 	return nil
 }
+
+// SystemSuspend triggers a system reboot.
```
gibmat: Comment is incorrect. ;)

stgraber: Oops :)

PR #901, opened 2026-02-17
Fix new linter warnings

gibmat:
Still a couple to do, but got the easy ones.

### DIFF
--- a/incus-osd/internal/storage/discard.go
+++ b/incus-osd/internal/storage/discard.go
@@ -24,7 +24,7 @@ func IsBlockdev(fm os.FileMode) bool {
 // An offset can be specified to only reset a part of a device.
 func ClearBlock(blockPath string, blockOffset int64) error {
 	// Open the block device for checking.
-	fd, err := os.OpenFile(blockPath, os.O_RDWR, 0o644)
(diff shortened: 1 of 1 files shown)


stgraber:
@gibmat looks like there's more :)

gibmat:
> @gibmat looks like there's more :)

Yeah, the remaining ones (G703, G704, G706) were too annoying to immediately fix. I'm tempted to just outright ignore them for now, since the instances I've looked at aren't actually issues, or try to do something like in the incus repo where only new changes are run through the linter.

stgraber:
I'd prefer not to introduce the logic to only scan new code, but I'm definitely happy to get rid of some of the gosec tests as they are a bit over the top sometimes.

breml:
I hit some of the gosec ones myself (also in other repos) and it looks, there has been a regression, which caused gosec to be to aggresive. Should be fixed in the next version of golangci-lint scheduled for release this week.

stgraber:
Dropped the gosec fixes from this branch and turned into configuration tweaks to golangci-lint instead, also re-enabling nolintlint to catch any leftover nolint entries.

stgraber (approved):

PR #905, opened 2026-02-19
Implement application debug API

### LINKED ISSUE #874: Implement an application debug API
Applications should be able to expose a function receiving a request struct and returning an application/octet-stream chunk of data for debugging purposes.

That would be slightly different from the more structured OS-level debug API (/1.0/debug) because we likely just want a consistent endpoint (/1.0/applications/NAME/:get-debug) that we POST a request to and not need a bunch of application-specific routes.

stgraber:
Closes #874

### DIFF
--- a/doc/reference/applications/incus.md
+++ b/doc/reference/applications/incus.md
@@ -30,3 +30,15 @@ Two additional applications exist which extend the main Incus application:
 
 * `incus-ceph`: Adds [Ceph](../services/ceph.md) client support
 * `incus-linstor`: Adds [Linstor](../services/linstor.md) satellite support
+
+## Debugging
+
+The Incus application supports the following debug actions:
+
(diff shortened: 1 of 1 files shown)


incus-osd/internal/rest/api_applications.go
```diff
@@ -209,6 +209,75 @@ func (s *Server) apiApplicationsEndpoint(w http.ResponseWriter, r *http.Request)
...
+	app, err := applications.Load(r.Context(), s.state, name)
+	if err != nil {
+		w.Header().Set("Content-Type", "application/json")
+
+		_ = response.InternalError(err).Render(w)
+
+		return
+	}
+
+	// Trigger the restart.
```
gibmat: Copy-n-pasta strikes again! ;)

gibmat (approved):

incus-osd/internal/rest/api_applications.go
stgraber: yup :)

PR #907, opened 2026-02-19
Add Netbird support

### LINKED ISSUE #219: Add netbird support
[Netbird](https://netbird.io/) is an alternative to Tailscale ([issue to include support](https://github.com/lxc/incus-os/issues/51)) which provides essentially the same functionality of creating an overlay network based on wireguard with SSO and MFA. It is also quite popular but, unlike tailscale, both the client and the management server are open source.

It would be nice to be able to configure it through the IncusOS network API to provide a secure wireguard overlay network between the hosts and for remote access.

jukoby:
Closes #219 

### Quick Summary of the choice of parameters

Always used:
`--no-browser`

Not used:
`--disable-auto-connect` managed by `services/service_netbird.go`
`--setup-key-file`
`--network-monitor` false by default
`--daemon-addr` is set in the service
`--service`
all SSH flags

I would love to have some input if the following ones should be added because I am unsure:
`--preshared-key` sets WireGuard PreSharedKey property. If set, then only peers that have the same key can communicate.
`--interface-name`
`--wireguard-port`
`--hostname` Does only work when connecting for the first time, can be changed in the admin ui
`--mtu`

Maybe it would make sense to always set these, would also love some input here:
 `--extra-iface-blacklist` Extra list of default interfaces to ignore for listening.
`--dns-router-interval` DNS route update interval (default 1m0s)

### DIFF
--- a/app-build/applications.json
+++ b/app-build/applications.json
@@ -171,6 +171,9 @@
     "netbird": {
         "version": "v0.64.6",
         "repo": "https://github.com/netbirdio/netbird.git",
+        "patches": [
+            "../../patches/netbird-0001-fix-go-1.26.patch"
+        ],
         "build_targets": [
             [
                 "-o",
--- a/patches/netbird-0001-fix-go-1.26.patch
+++ b/patches/netbird-0001-fix-go-1.26.patch
@@ -1,0 +1,35 @@
+diff --git a/go.mod b/go.mod
+index 801d5248..f605c9c0 100644
+--- a/go.mod
++++ b/go.mod
+@@ -1,8 +1,6 @@
+ module github.com/netbirdio/netbird
+ 
+-go 1.25
+-
+-toolchain go1.25.5
++go 1.25.5
+ 
+ require (
+ 	cunicu.li/go-rosenpass v0.4.0
+@@ -123,7 +121,7 @@ require (
+ 	gorm.io/driver/postgres v1.5.7
+ 	gorm.io/driver/sqlite v1.5.7
+ 	gorm.io/gorm v1.25.12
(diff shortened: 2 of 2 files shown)


mkosi.images/base/mkosi.extra/usr/lib/systemd/system/netbird.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=NetBird mesh network client
+
+[Service]
+StartLimitInterval=5
+StartLimitBurst=10
+ExecStart=/usr/bin/netbird service run --log-level info --daemon-addr unix:///var/run/netbird.sock --log-file /var/log/netbird/client.log
+StandardOutput=file:/var/log/netbird/netbird.out
+StandardError=file:/var/log/netbird/netbird.err
```
stgraber: We should drop all the /var/log/netbird stuff.
IncusOS entirely relies on the systemd journal for logging and actively tries to disable all other logging mechanisms.

That's because we have added support for streaming the journal over remote syslog and also have rotation/cleanup policies for the journal to handle low disk space situations.

stgraber:
Change looks great!

There's a conflict in `app-build/build-applications.py` that needs addressing as well as my comments on the systemd unit, but other than that, this looks very good and could be merged so we can get a testing build out for you to validate with an actual Netbird server.

jukoby:
Fixed the systemd unit and the merge conflict, thank you for the feedback!

jukoby:
The reason the build is failing seems to be [this issue in netbird](https://github.com/netbirdio/netbird/issues/5290).
Is there a way around this or should we wait until [the fix is merged](https://github.com/netbirdio/netbird/pull/5447)?

PR #912, opened 2026-02-25
tailscale: Add support for tailscale services in serve command

### LINKED ISSUE #911: Add support for tailscale services?
Tailscale since the fall update has support for [services](https://tailscale.com/docs/features/tailscale-services) -- which I feel would be a good usecase of having the nodes in a cluster all exposing on that service to provide fairly-simple HA for the cluster's control plane.

It seems this feature [isn't implemented in headscale yet](https://github.com/juanfont/headscale/issues/2845) so if that is a deal-breaker, please let me know. Otherwise I'll look into opening a PR sometime soon, as I don't anticipate it being a very big change to add support (Basically just need to add a new config field, and change the `tailscale serve` command to pass it in)

AmandaCameron:
I'm fairly sure this doesn't require a change in the tailscale binary's build flags, but some advice on how to test the incusos image on a non-local incus cluster would be appreciated so I can verify that's the case. (I've got an incusos instance in a local libvirt, and my incusos cluster on a trio of NUCs)

Closes #911

### DIFF
--- a/doc/.wordlist.txt
+++ b/doc/.wordlist.txt
@@ -36,6 +36,7 @@ FuturFusion
 GiB
 Github
 GPUs
+Headscale
 homelab
 hotfix
 HP


AmandaCameron:
Well that's embarassing. I could have sworn I checked for syntax errors before committing, fixed now.

AmandaCameron:
The docs check is unhappy about the use of "Headscale" -- not sure if that's something we even want mentioned in the docs, so I didn't immediately add it to the dictionary.

AmandaCameron:
Terribly sorry about the commit spam. having A Day it seems. I can rebase them all together if they look good. otherwise

stgraber:
Headscale needs to be added to the wordlist in doc/.wordlist.txt

AmandaCameron:
Added. I can test this against my own tailscale account to make sure it actually functions without any changes needed to the tailscale build tags, I just don't know how to take the iso from flasher-image and put it into a remote incus instance, as everything I've seen seems to assume it shares a fs with the client?

PR #915, opened 2026-02-26
Use full device IDs during install

### LINKED ISSUE #866: During install, report drives by ID in messages
Currently, during install we report drives via their short name (ie, `/dev/sda`), but we should switch to using `/dev/disk/by-id/...`. This will make it clearer exactly what drive is being used for the install, as well as help fix some issues with the tests where scsi ordering occasionally differs and causes `/dev/sda` to become `/dev/sdb` (or vice versa).

gibmat:
Use full "by-id" device IDs in install messages and errors. This should help make it clearer exactly what source and target devices are being used as well as eliminate occasional test failures caused by differences in the enumeration of scsi devices by the kernel.

Closes #866

### DIFF
--- a/incus-osd/tests/incusos_tests/incus_test_vm/__init__.py
+++ b/incus-osd/tests/incusos_tests/incus_test_vm/__init__.py
@@ -86,7 +86,7 @@ class IncusTestVM:
         else:
             subprocess.run(["incus", "stop", self.vm_name], capture_output=True, check=True, timeout=timeout)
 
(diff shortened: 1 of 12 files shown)


stgraber:
@gibmat you've upset revive :)
We may need to bump the max complexity config again, or find a way to simplify that function.

gibmat:
> @gibmat you've upset revive :) We may need to bump the max complexity config again, or find a way to simplify that function.

Just saw that. :grinning:

PR #916, opened 2026-02-27
Improve recovery prompt

### LINKED ISSUE #859: Improve recovery passphrase prompt
At the moment, the encryption recovery passphrase prompt doesn't re-prompt if an incorrect recovery passphrase is entered, and no other feedback is provided. We should try to make this more informative and user-friendly.

gibmat:
Adjust the encryption recovery boot prompt to re-prompt until a correct passphrase is entered.

Closes #859

### DIFF
--- a/mkosi.packages/incusos-initrd-utils/initrd-multipath-partition.sh
+++ b/mkosi.packages/incusos-initrd-utils/initrd-multipath-partition.sh
@@ -27,12 +27,12 @@ for MP in $(dmsetup ls --target multipath | cut -f1); do
 
     # Attempt to unlock swap and root partitions.
     if [ -e "/dev/mapper/${MP}-part9" ]; then
-        # Unlock swap and manually activate since it's not automatically picked up by systemd.
-        systemd-cryptsetup attach swap "/dev/mapper/${MP}-part9"
-        swapon /dev/mapper/swap
-
         # Unlock root partition, which will be automatically detected and then mounted.
-        systemd-cryptsetup attach root "/dev/mapper/${MP}-part10"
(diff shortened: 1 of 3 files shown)


mkosi.packages/incusos-initrd-utils/initrd-multipath-partition.sh
```diff
@@ -27,12 +27,12 @@ for MP in $(dmsetup ls --target multipath | cut -f1); do
 
     # Attempt to unlock swap and root partitions.
     if [ -e "/dev/mapper/${MP}-part9" ]; then
+        # Unlock root partition, which will be automatically detected and then mounted.
+        systemd-cryptsetup attach "root" "/dev/mapper/${MP}-part10" "" "tpm2-device=auto,tpm2-measure-pcr=yes,tries=0"
```
gibmat, line 31: Switch to unlocking the root partition first, so the recovery prompt is the same regardless of multipath being enabled or not.

mkosi.packages/incusos-initrd-utils/systemd-cryptsetup.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/systemd-cryptsetup attach 'root' '/dev/gpt-auto-root-luks' '' 'tpm2-device=auto,tpm2-measure-pcr=yes,tries=0'
```
gibmat, line 3: Ideally, `systemd-gpt-auto-generator` and `systemd-cryptsetup-generator` would pickup kernel command arguments to adjust the options passed to cryptsetup, but that doesn't seem to work. For IncusOS, we can safely override the generated command, since our root LUKS setup will always be the same.

PR #937, opened 2026-03-06
Switch to downloading pre-compiled UI and Migration Manager assets

### LINKED ISSUE #913: Use published UI artifacts for Migration Manager and Operations Center
Now that Migration Manager and Operations Center are publishing pre-built UI tar archives with each release, we should remove the UI build logic from IncusOS and instead simply fetch and extract the UI tar archives directly.

gibmat:
This eliminates the need to install yarn (and all its dependencies) during a build of IncusOS.

Also removed logic for building the Migration Manager worker x86_64 image in favor of downloading the asset from the Migration Manager release on github.

Closes #913

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,14 +43,12 @@ jobs:
             binutils \
             debian-archive-keyring \
             devscripts \
-            libnbd-dev \
             make \
             mtools \
             parted \
             pkgconf \
             pipx \
-            qemu-utils \
-            yarnpkg
+            qemu-utils
 
       - name: Setup mkosi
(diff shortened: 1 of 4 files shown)


stgraber:
@gibmat looks like extractall tries to fit the whole thing in memory and then gets killed by the kernel out of memory killer ;)

gibmat:
> @gibmat looks like extractall tries to fit the whole thing in memory and then gets killed by the kernel out of memory killer ;)

Added some chunking, and verified it works when limited to 500MB of RAM. :)

stgraber (approved):

PR #939, opened 2026-03-06
Install logic fixes

### LINKED ISSUE #925: Two tests failing after multipath changes
Two tests are failing after the multipath changes were merged:

* `TestInstallNoSeedReadonlyImage/img`
* `TestInstallDontRemoveInstallMedia/iso`

It appears that expected `/dev/disk/by-partlabel/` symlinks are different, likely caused by the multipath service delaying initrd startup while udev settles.

gibmat:
* Improve the detection of a read-only source device
* Use a custom `IncusOSInstallComplete` UEFI variable to indicate when an install has completed successfully. This will be cleared by `incus-osd` on first boot, but also provides a more robust way to detect if a user accidentally reboots into an install media without relying on detecting symlinks.

Closes #925

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -103,8 +103,7 @@ jobs:
             mtools \
             pkgconf \
             pipx \
-            qemu-utils \
-            yarnpkg
+            qemu-utils
 
       - name: Setup Incus
         run: |


gibmat:
Something weird is happening that's causing the `incus exec test-incus-os -- curl --unix-socket /run/incus-os/unix.socket http://localhost/1.0/system/update/:check -X POST` call to fail; it seems to work fine locally....

PR #941, opened 2026-03-08
base: Add drbd patch for 6.19 kernels

stgraber:


stgraber:
Self-merging as this fix is pretty obvious (taken from upstream) and fixed a test and build failure.

PR #942, opened 2026-03-09
Fix wiping a LUKS-encrypted drive

gibmat:
Make sure the LUKS mapping is properly closed and remove the encryption key before wiping a LUKS-encrypted drive. Added tests to exercise this part of the API.

Also dropped drdb patch that's been included in that package's latest release.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,8 +77,8 @@ jobs:
     timeout-minutes: 120
     runs-on:
       - self-hosted
-      - cpu-4
-      - mem-4G
+      - cpu-8
+      - mem-16G
       - disk-100G
       - arch-amd64
       - image-debian-13


.github/workflows/build.yml
```diff
@@ -19,8 +19,8 @@ jobs:
     timeout-minutes: 90
     runs-on:
       - self-hosted
-      - cpu-4
-      - mem-4G
+      - cpu-8
+      - mem-16G
```
stgraber: Wrong workflow, it's the test one that's failing :)

gibmat: Argh.... :)

incus-osd/internal/storage/storage.go
```diff
@@ -963,6 +963,9 @@ func isBootDevice(ctx context.Context, deviceName string, bootDevice string) boo
 
 // WipeDrive will wipe all data on the given drive, unless it is the boot device,
 // a remote device, or currently a member of a storage pool.
+//
+// If the deivce is a LUKS-encrypted drive, first attempt to close the LUKS mapping
```
stgraber: typo, should be `device` :)

PR #953, opened 2026-03-12
Update LUKS logic to use an internal recovery key

gibmat:
When generating the initial encryption recovery key, create an internal, IncusOS-only one in addition to the one exposed via API. This gives us a fallback when dealing with LUKS volumes if the TPM state is incorrect, and also eliminates the need to pass around an encryption key since we can just read it from disk.

Logic is also included to migrate existing IncusOS systems via a one-time migration of any existing encryption keys.

### DIFF
--- a/incus-osd/internal/secureboot/pcrs.go
+++ b/incus-osd/internal/secureboot/pcrs.go
@@ -21,6 +21,7 @@ import (
 	"github.com/lxc/incus/v6/shared/subprocess"
 	"github.com/smallstep/pkcs7"
 
+	"github.com/lxc/incus-os/incus-osd/internal/state"
 	"github.com/lxc/incus-os/incus-osd/internal/util"
 )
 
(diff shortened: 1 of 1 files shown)


incus-osd/internal/systemd/cryptenroll.go
```diff
@@ -16,31 +16,79 @@ import (
 	"github.com/lxc/incus-os/incus-osd/internal/util"
 )
 
-// GenerateRecoveryKey utilizes systemd-cryptenroll to generate a recovery key for the
-// root and swap LUKS volumes. Depends on an existing tpm2-backed key being enrolled and accessible.
-func GenerateRecoveryKey(ctx context.Context, s *state.State) error {
+// GenerateRecoveryKeys utilizes systemd-cryptenroll to generate recovery keys for the
+// root and swap LUKS volumes. The logic depends on an existing tpm2-backed key being
+// enrolled and accessible, which is should true on first boot.
```
stgraber: `is should true` :)

Is it supposed to be `is true on first boot` or `should be true on first boot`?

incus-osd/cmd/incus-osd/main.go
```diff
@@ -384,7 +384,7 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error { //nolint:r
 	if len(s.System.Security.Config.EncryptionRecoveryKeys) == 0 {
```
stgraber, line 384: Looking at this PR with the migration code doing most of what GenerateRecoveryKeys does, wouldn't it have been simpler to change this check for one checking whether /var/lib/incus-os/recovery.root.key exists and if not, run through the GenerateRecoveryKeys logic, just having it skip the user visible recovery key if len(s.System.Security.Config.EncryptionRecoveryKeys) > 0?

Feels like that would let us use the first-boot code to handle the migration too rather than have duplicated logic, or was that too difficult/fragile for some reason?

incus-osd/internal/systemd/cryptenroll.go
gibmat: Grammar is hard :)

incus-osd/cmd/incus-osd/main.go
gibmat: Hmm, yeah, it is possible to consolidate everything into `GenerateRecoveryKeys()`. It turned out not to be as complicated as I was expecting, which is why I didn't do it initially.

PR #965, opened 2026-03-15
preserve zd* /dev/zvol symlinks for LINSTOR

chribro88:
preserve the OpenZFS-managed `/dev/zvol` symlinks for zd* devices so LINSTOR can use the standard `/dev/zvol/<pool>/<identifier>` path


chribro88:
from LINSTOR error report:
```bash
Error message: Failed to access DRBD super-block of volume incus-volume-833a540ab494414fb35ac494f29f54c1/0
```
```bash
Caused by:
Class name: FileNotFoundException
```
```bash
Error message: /dev/zvol/tank/linstor/incus-volume-833a540ab494414fb35ac494f29f54c1_00000 (No such file or directory)
```

and zd* devices on the nodes:
```bash
# linstor physical-storage list
╭────────────────────────────────────────────────────────────────────────╮
┊ Size        ┊ Rotational ┊ Nodes                                       ┊
╞════════════════════════════════════════════════════════════════════════╡
┊ 10737418240 ┊ False      ┊ server02[/dev/zd0,/dev/zd16]                ┊
┊             ┊            ┊ server01[/dev/zd0,/dev/zd16,/dev/zd32]      ┊
┊             ┊            ┊ server03[/dev/zd0]                          ┊
╞┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄┄╡
┊ 10739761152 ┊ False      ┊ server01[/dev/zd48]                         ┊
┊             ┊            ┊ server03[/dev/zd16]                         ┊
╰────────────────────────────────────────────────────────────────────────╯
```

stgraber:
I don't think that this change would prevent udev from creating other symlinks outside of /dev/zvol which is what we were trying to do after having issues of /dev/disk/by-partlabel/ and such getting symlinks to instance volumes rather than physical disks.

But because ZFS doesn't do partition probing unless `volmode` is specifically set to `full` and Incus never does that (only uses `none` or `dev`), then this should be safe.

stgraber (approved):

PR #972, opened 2026-03-17
Make `manifests` package public

gibmat:


stgraber:
@gibmat just a tiny nit but given the package is named `manifests`, I think we can do away with the `Manifests` suffixes in all the exported functions.

So we'd just have `manifests.Read`, `manifests.Diff`, `manifests.Write`, ...

I'm actually surprised that golangci-lint didn't yell at us about that :)

gibmat:
> @gibmat just a tiny nit but given the package is named `manifests`, I think we can do away with the `Manifests` suffixes in all the exported functions.

Sure, I'll make that update real quick and push another commit.

PR #974, opened 2026-03-17
Implement network configuration rollback

### LINKED ISSUE #860: Add network configuration rollback
When configuring the network, a mistake can knock IncusOS off the network and there's currently no easy way to recover from this. We should add an option when setting the network configuration to automatically rollback after a configured period of time unless canceled. We'd then add some sort of "commit" or "confirm" API endpoint that would make the new configuration permanent.

Because of the complexity of network configurations, there's no way to automatically detect if a configuration is "good", either from the IncusOS side (ie, DNS checks, pings, etc) or from the client side (since the IP address(es) may change, etc). The user will need to run two different commands:

    incus admin os system network edit

Then, after confirming that IncusOS appears correctly on the network:

    incus admin os system network commit

gibmat:
Closes #860

### DIFF
--- a/incus-osd/tests/api-tests.py
+++ b/incus-osd/tests/api-tests.py
@@ -70,7 +70,7 @@ num_pass = 0
 num_fail = 0
 
 # Run the tests
-with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
+with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
     tests = IncusOSTests(prior_image_img, current_image_img, current_image_iso)
     futures = {executor.submit(fn, image): name for name,fn,image in tests.GetTests()}
 


incus-osd/internal/rest/api_system_network.go
```diff
@@ -116,24 +118,86 @@ func (s *Server) apiSystemNetwork(w http.ResponseWriter, r *http.Request) {
...
+				for time.Since(startTime) < confirmationTimeout {
+					// If NetworkConfigurationPending is false, that means the user has confirmed
+					// the changes, so there's no need to restore the prior network configuration.
+					if !s.state.NetworkConfigurationPending {
+						return
+					}
+
+					// Sleep five seconds before checking again.
+					time.Sleep(5 * time.Second)
+				}
```
stgraber: That's a bit racy though, I could:
 - Perform a config change with a revert set to 5min
 - Confirm it
 - Within 5s, send another config change with a revert set to 5min
 - Have the go routine of the previous change kick in and revert to the old-old-config
 - Have the go routine handling the second change then exit thinking the user confirmed the change

It's not super likely to happen, you'd need someone who's making two consecutive very small edits, but it may be possible when tweaking something like VLANs.

Using a channel for confirmation would allow for using a `select` statement waiting on either user confirmation OR a timeout, eliminating the need for the sleep 5s loop and the race condition in the process.

incus-osd/internal/rest/api_system_network.go
```diff
@@ -219,3 +219,40 @@ func applyNetworkConfiguration(ctx context.Context, s *state.State, networkCfg *
...
+//	consumes:
+//	  - application/json
+//	produces:
+//	  - application/json
+//	responses:
+//	  "200":
+//	    $ref: "#/responses/EmptySyncResponse"
+//	  "400":
+//	    $ref: "#/responses/BadRequest"
+func (s *Server) apiSystemNetworkConfirm(w http.ResponseWriter, r *http.Request) {
```
stgraber, line 247: The function should probably BadRequest if no confirmation timeout was defined.

I also think we should clear the ConfirmationTimeout field from the configuration if it was set. That's to avoid a scenario where a tool like Operations Center would then do a GET+PUT and trigger the confirmation logic.

(Operations Center currently has a number of codepaths which can lead to "pointless" network configuration changes where the current value is effectively read and written back as-is. If a confirmation_timeout was part of that value it would trigger a rollback x minutes later every time it does such a pointless update)

PR #989, opened 2026-03-25
Version applications

### LINKED ISSUE #347: Version the application updates
Currently we only store one version of the application, the current running one.
This may or may not line up with the version of the OS.

The behavior is generally fine as we want to allow for the applications to update and restart independently of the OS, allowing for bugfixes and even new features to be rolled out without needing a restart.

However the issue with the current approach is that we have no way to rollback a bad application update, unlike what we can do with the OS where you can boot an older version to get out of a bad update.

To fix this, we should:
 - Keep the previous version of an application around (effectively giving us the same A/B mechanism as the OS)
 - Have an API mechanism to roll back onto the previous version (/1.0/applications/NAME/rollback or something like that)
 - Make it so booting the backup OS version also loads up the backup application versions. The OS and application versions may not line up, that's fine, but if booting in what's effectively a recovery mode, we should also load the older application in case the issue the user is running into is application rather than OS related.

gibmat:
Closes #347

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -2,7 +2,7 @@ definitions:
     ACMEChallengeType:
         title: ACMEChallengeType represents challenge types for ACME configuration.
         type: string
-        x-go-package: github.com/FuturFusion/operations-center/shared/api/system
+        x-go-package: github.com/FuturFusion/migration-manager/shared/api
     Access:
         items:
             $ref: '#/definitions/AccessEntry'
@@ -3540,6 +3540,11 @@ definitions:
                 x-go-name: SourceSpecificID
             type:
                 $ref: '#/definitions/NetworkType'
+                description: The network type
+                example: bridge
+                readOnly: true
+                type: string
+                x-go-name: Type
             uuid:
                 description: UUID of the network.
                 format: uuid
@@ -6993,11 +6998,11 @@ definitions:
                     type: string
                 description: Server configuration map (refer to doc/server.md)
                 example:
-                    core.https_address: :8443
+                    core.https_address: :6443
                 type: object
                 x-go-name: Config
         type: object
-        x-go-package: github.com/lxc/incus/v6/shared/api
+        x-go-package: github.com/FuturFusion/migration-manager/shared/api
     ServerStorageDriverInfo:
         description: ServerStorageDriverInfo represents the read-only info about a storage driver
         properties:
@@ -8459,10 +8464,13 @@ definitions:
         properties:
             status:
                 $ref: '#/definitions/WarningStatus'
+                description: Status of the warning (new, acknowledged, or resolved)
+                example: new
                 type: string
-        title: WarningPut represents the modifiable fields of a warning.
+                x-go-name: Status
+        title: WarningPut represents configurable properties of a warning.
         type: object
-        x-go-package: github.com/lxc/incus/v6/shared/api
+        x-go-package: github.com/FuturFusion/migration-manager/shared/api
     WarningScope:
         properties:
             entity:
@@ -8653,8 +8661,11 @@ paths:
                                 example:
                                     config: {}
                                     state:
+                                        available_versions:
+                                            - "202511041601"
+                                            - "202511041800"
                                         initialized: true
-                                        version: "202511041601"
+                                        version: "202511041800"
                                 type: json
                             status:
                                 description: Status description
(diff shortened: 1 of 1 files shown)


incus-osd/cmd/incus-osd/main.go
```diff
@@ -513,6 +513,55 @@ func startup(ctx context.Context, s *state.State, t *tui.TUI) error { //nolint:r
 		}
 	}
 
+	// Migrate existing sysext images under /var/lib/extensions/ to /var/lib/incus-os-extensions/.
```
gibmat: I've tested this as best I can mimicking what the images provider should do when providing an existing IncusOS install with updated, versioned application images. We'll want to quickly test the updates when published to the testing channel and verify that post migration, only symlinks exist under `/var/lib/extensions/`, and the most recent versioned application image exists under `/var/lib/incus-os-extensions/`.

On an existing IncusOS system, the applications should "update" although `systemd-sysext` will actually layer the new versioned image over the existing unversioned image. Upon rebooting, this migration logic should take care of cleaning up old unversioned image(s) and switching to the new versioned application image(s).

gibmat (commented):
Alright, I think this should be ready for review. I've kicked the tires pretty well, and everything should be working as expected:

* Migrating existing IncusOS installs
* Switching application versions, both automatically when booting into the backup A/B image, as well as via API
* Updated manifest and changelog generation to handle versioned application images

stgraber:
@gibmat any idea what's going on with the test failure?

gibmat:
> @gibmat any idea what's going on with the test failure?

Not sure right off hand; for some reason the actual systemd services failed to start and I don't know if that's because the sysext failed to load or not. I haven't been able to immediately reproduce this on my end yet.

incus-osd/internal/state/struct.go
```diff
@@ -18,10 +18,11 @@ type SecureBoot struct {
...
 // OS represents the current OS image state.
 type OS struct {
-	Name           string `json:"name"`
-	RunningRelease string `json:"running_release"`
-	NextRelease    string `json:"next_release"`
-	SuccessfulBoot bool   `json:"successful_boot"`
+	Name             string `json:"name"`
+	RunningRelease   string `json:"running_release"`
+	NextRelease      string `json:"next_release"`
+	PriorBootRelease string `json:"prior_boot_release"`
```
stgraber: Hmm, I'm not convinced about the naming here.
PriorBootRelease makes it sound to me like the boot -1 version, but we may have booted twice from the backup image or may have booted multiple times from the current image.

Maybe BackupRelease would be easier to understand?

incus-osd/manifests/manifests.go
```diff
@@ -215,12 +215,20 @@ func Read(root string, manifests []string) (map[string]IncusOSManifest, error) {
 	for _, manifestName := range manifests {
 		var m IncusOSManifest
 
-		file := filepath.Join(root, manifestName+".manifest")
+		// Every mkosi manifest will include the version in its filename, except for the base manifest.
```
stgraber: Okay, so we'll need to also release a new Operations Center to handle this then as otherwise any codepath that generates changelogs will be broken.

incus-osd/internal/rest/api_applications.go
```diff
@@ -661,3 +662,111 @@ func (s *Server) apiApplicationsRemove(w http.ResponseWriter, r *http.Request) {
 
 	_ = response.EmptySyncResponse.Render(w)
 }
+
+// swagger:operation POST /1.0/applications/{name}/:set-version applications applications_post_set_version
```
stgraber: Would `switch-version` be clearer?

stgraber:
@gibmat so why do we have to use the versioned mkosi extensions?

That seems to be causing a lot of changes that I'm not sure I understand why we need them at all. We only ever have one version of a given application within a given update. An update already has a version string, so why can't we just use them as-is and just keep track of the version at download time within incus-osd?

With the change to all the mkosi filenames, we've got a ripple effect in needing the publication logic changed on our servers, mirroring patterns at some customers may need updating, Operations Center will break and need a new release and I'm also not sure how exactly an older IncusOS is supposed to be able to download a new application prior to it having updated its OS and rebooted.

It feels to me like this PR would be a whole lot simpler if it wasn't for this change, so I'm trying to understand why it was needed.

incus-osd/internal/rest/api_applications.go
gibmat: Yeah, that's nicer.

incus-osd/internal/state/struct.go
gibmat: It does track the prior (-1) boot release. If the system boots twice in a row to the same OS version, `PriorBootRelease == RunningRelease`. This is important because we need to detect the first time we change OS versions so application versions can be automatically switched, but then on reboots to the same OS version we don't want to overwrite application version changes that the user may have made. (For example, the user boots to the backup IncusOS version, but still wants to use the latest available Incus application.)

In thinking about this, I realized I can get the equivalent information from `journalctl`, so I'll switch to that and avoid introducing another state field. :)

stgraber: Ah, okay, make sense. I was just worried we'd start exposing that to users and cause a bunch of confusions as we're now juggling a LOT of versions :)

gibmat:
@stgraber v2 is now ready. I've re-tested things and I think this is ready once again for a review.

PR #1000, opened 2026-03-31
incus-osd: Use modern YAML API but v2 rendering

stgraber:


gibmat:
Whoo! #1000!

stgraber:
That went by quickly :)

PR #1013, opened 2026-04-02
Add application and OS only update checks via REST API

### LINKED ISSUE #999: Add application update check API
> @gibmat just FYI, but we've been talking with @breml about implementing a `/os/1.0/applications/NAME/:check-update` type API so it's possible to run a system with a `never` update frequency and then update applications or OS independently.
> 
> Just worth keeping in mind as we work on refactoring some of that update code and adding the support for multiple application versions to IncusOS.
> 
> API wise, I think we'll just need:
>  - POST /1.0/applications/NAME/:check-update
>  - Tweak to POST /1.0/system/update/:check to allow restricting the update check to just the OS and not applications on it 

 _Originally posted by @stgraber in [#561](https://github.com/FuturFusion/operations-center/issues/561#issuecomment-3738672937)_

gibmat:
Closes #999

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -2,7 +2,7 @@ definitions:
     ACMEChallengeType:
         title: ACMEChallengeType represents challenge types for ACME configuration.
         type: string
-        x-go-package: github.com/FuturFusion/migration-manager/shared/api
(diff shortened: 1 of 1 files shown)


gibmat:
@breml This adds an API to IncusOS to update individual applications and adds an optional `os_only` parameter to the system-level update check.

PR #1026, opened 2026-04-07
doc/reference/system/network: Add section on valid `required_for_online` values

gibmat:


stgraber (approved):

gibmat:
Huh...

> (reference/system/network: line   61) broken    https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#RequiredFamilyForOnline= - 418 Client Error: I'm a teapot for url: https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html

stgraber:
@gibmat you'll want to put it in the exclusion list, that's an anti-scrapper thing

PR #1043, opened 2026-04-17
Add netdev_max_backlog, tcp_mtu_probing, and ring buffer settings for high-speed NICs

### LINKED ISSUE #1012: Add netdev_max_backlog, tcp_mtu_probing, and ring buffer settings for high-speed NICs
# IncusOS Feature Request: High-Speed Network Tunables

**Target repo:** https://github.com/lxc/incus-os/issues
**Related:** Issue #730 (TCP tuning), PR #754 (system/kernel API), Issue #630 (system/kernel design)

---

## Description

The `/1.0/system/kernel` API (added in PR #754) covers TCP buffer sizing and congestion control, which is great. However, deploying IncusOS on servers with 100Gb NICs and jumbo frames (MTU 9000) reveals three gaps where stock kernel defaults cause measurable performance loss:

### 1. `net.core.netdev_max_backlog` (kernel API addition)

**Current:** Stock default is 1000.
**Want:** 30000+ for 100Gb NICs.
**Why:** At 100Gb line rate, the NIC can deliver ~8.3 million packets/second (at MTU 9000) or ~148 million pps (at 64 bytes). The per-CPU backlog queue fills in microseconds at stock depth. Under burst from multiple VMs simultaneously using the network, this causes silent packet drops before the kernel even processes them. This is the single most impactful tunable for high-speed NIC hypervisors.

**Proposed API field:**
```go
type SystemKernelConfigNetwork struct {
    // ... existing fields ...
    NetdevMaxBacklog int `json:"netdev_max_backlog,omitempty" yaml:"netdev_max_backlog,omitempty"`
}
```

### 2. `net.ipv4.tcp_mtu_probing` (kernel API addition)

**Current:** Stock default is 0 (disabled).
**Want:** 1 (enabled, probe on ICMP blackhole detection).
**Why:** Required for Path MTU Discovery to work reliably with jumbo frames (MTU 9000). Without it, PMTUD blackholes silently kill throughput on any path that doesn't support the full MTU. This is a boolean-level setting with zero downside — the kernel only activates probing when it detects a blackhole.

**Proposed API field:**
```go
type SystemKernelConfigNetwork struct {
    // ... existing fields ...
    TCPMTUProbing bool `json:"tcp_mtu_probing,omitempty" yaml:"tcp_mtu_probing,omitempty"`
}
```

### 3. Ring buffer sizing (network API addition)

**Current:** The `SystemNetworkEthernet` struct exposes offload toggles (GRO, GSO, TSO) but not ring buffer sizes.
**Want:** Fields to set Rx, Tx, and RxJumbo ring buffer sizes per interface/bond.
**Why:** Larger ring buffers absorb burst traffic from multiple VMs sharing the physical NIC. The NIC driver reports maximum supported sizes via ethtool; the ideal setting is "set all to max." For reference, Broadcom 100Gb NICs support Rx=2047, Tx=2047, RxJumbo=8191. Stock defaults are typically much lower.

systemd-networkd already supports this via `.link` files (`RxBufferSize=`, `TxBufferSize=`, `RxJumboBufferSize=`), and IncusOS's `generateEthernet()` function already maps struct fields to .link directives — so the plumbing is in place.

**Proposed API fields:**
```go
type SystemNetworkEthernet struct {
    // ... existing fields ...
    RxBufferSize    *int `json:"rx_buffer_size,omitempty" yaml:"rx_buffer_size,omitempty"`
    TxBufferSize    *int `json:"tx_buffer_size,omitempty" yaml:"tx_buffer_size,omitempty"`
    RxJumboBufferSize *int `json:"rx_jumbo_buffer_size,omitempty" yaml:"rx_jumbo_buffer_size,omitempty"`
    RxMiniBufferSize  *int `json:"rx_mini_buffer_size,omitempty" yaml:"rx_mini_buffer_size,omitempty"`
}
```

Or alternatively, a single `ring_buffers: max` option that auto-discovers and sets all ring buffers to their hardware maximum — which is what most 100Gb deployments want.

## Environment

- Hardware: HP ProLiant Gen10 with Broadcom 100Gb NICs (ens1f0np0/ens1f1np1 in LACP bond)
- MTU: 9000 (jumbo frames)
- Workload: VM hypervisor running network-intensive workloads (NFS, Ceph replication, VM traffic on multiple VLANs)
- IncusOS deployed on 3 nodes initially, scaling to 17

## Impact

These three settings are universally beneficial for any IncusOS deployment on 10Gb+ hardware. They're non-destructive (no downside to higher backlog or enabled PMTUD), align with the "do the right thing out of the box" philosophy, and address the most common performance gap when moving from a tuned Linux hypervisor to IncusOS.

sharathsivakumar:
- support setting ring buffer sizes per interface
- introduce TCPMTUProbing in SystemKernelConfigNetwork
- NetdevMaxBacklog in SystemKernelConfigNetwork

Closes #1012

### DIFF
--- a/incus-osd/api/system_kernel.go
+++ b/incus-osd/api/system_kernel.go
@@ -18,8 +18,8 @@ type SystemKernelConfigNetwork struct {
 	BufferSize             int    `json:"buffer_size,omitempty"              yaml:"buffer_size,omitempty"`
 	QueuingDiscipline      string `json:"queuing_discipline,omitempty"       yaml:"queuing_discipline,omitempty"`
 	TCPCongestionAlgorithm string `json:"tcp_congestion_algorithm,omitempty" yaml:"tcp_congestion_algorithm,omitempty"`
-	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty" yaml:"netdev_max_backlog,omitempty"`
-	TCPMTUProbing          bool   `json:"tcp_mtu_probing,omitempty" yaml:"tcp_mtu_probing,omitempty"`
+	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty"       yaml:"netdev_max_backlog,omitempty"`
+	TCPMTUProbing          bool   `json:"tcp_mtu_probing,omitempty"          yaml:"tcp_mtu_probing,omitempty"`
 }
 
 // SystemKernelConfigPCI holds PCI-specific kernel configuration.
--- a/incus-osd/api/system_kernel.go
+++ b/incus-osd/api/system_kernel.go
@@ -18,6 +18,7 @@ type SystemKernelConfigNetwork struct {
 	BufferSize             int    `json:"buffer_size,omitempty"              yaml:"buffer_size,omitempty"`
 	QueuingDiscipline      string `json:"queuing_discipline,omitempty"       yaml:"queuing_discipline,omitempty"`
 	TCPCongestionAlgorithm string `json:"tcp_congestion_algorithm,omitempty" yaml:"tcp_congestion_algorithm,omitempty"`
+	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty" yaml:"netdev_max_backlog,omitempty"`
 }
 
 // SystemKernelConfigPCI holds PCI-specific kernel configuration.
--- a/incus-osd/api/system_kernel.go
+++ b/incus-osd/api/system_kernel.go
@@ -19,6 +19,7 @@ type SystemKernelConfigNetwork struct {
 	QueuingDiscipline      string `json:"queuing_discipline,omitempty"       yaml:"queuing_discipline,omitempty"`
 	TCPCongestionAlgorithm string `json:"tcp_congestion_algorithm,omitempty" yaml:"tcp_congestion_algorithm,omitempty"`
 	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty" yaml:"netdev_max_backlog,omitempty"`
+	TCPMTUProbing          bool   `json:"tcp_mtu_probing,omitempty" yaml:"tcp_mtu_probing,omitempty"`
 }
 
 // SystemKernelConfigPCI holds PCI-specific kernel configuration.
--- a/incus-osd/api/system_network.go
+++ b/incus-osd/api/system_network.go
@@ -98,6 +98,10 @@ type SystemNetworkEthernet struct {
 	WakeOnLAN              bool     `json:"wakeonlan,omitempty"                yaml:"wakeonlan,omitempty"`
 	WakeOnLANModes         []string `json:"wakeonlan_modes,omitempty"          yaml:"wakeonlan_modes,omitempty"`
 	WakeOnLANPassword      string   `json:"wakeonlan_password,omitempty"       yaml:"wakeonlan_password,omitempty"`
+	RxBufferSize           *int     `json:"rx_buffer_size,omitempty"           yaml:"rx_buffer_size,omitempty"`
+	TxBufferSize           *int     `json:"tx_buffer_size,omitempty"           yaml:"tx_buffer_size,omitempty"`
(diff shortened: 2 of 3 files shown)


incus-osd/api/system_network.go
```diff
@@ -98,6 +98,10 @@ type SystemNetworkEthernet struct {
 	WakeOnLAN              bool     `json:"wakeonlan,omitempty"                yaml:"wakeonlan,omitempty"`
 	WakeOnLANModes         []string `json:"wakeonlan_modes,omitempty"          yaml:"wakeonlan_modes,omitempty"`
 	WakeOnLANPassword      string   `json:"wakeonlan_password,omitempty"       yaml:"wakeonlan_password,omitempty"`
+	RxBufferSize           *int     `json:"rx_buffer_size,omitempty"           yaml:"rx_buffer_size,omitempty"`
+	TxBufferSize           *int     `json:"tx_buffer_size,omitempty"           yaml:"tx_buffer_size,omitempty"`
+	RxJumboBufferSize      *int     `json:"rx_jumbo_buffer_size,omitempty"     yaml:"rx_jumbo_buffer_size,omitempty"`
+	RxMiniBufferSize       *int     `json:"rx_mini_buffer_size,omitempty"      yaml:"rx_mini_buffer_size,omitempty"`
```
stgraber, line 104: Why are those pointers? None of the buffers can be 0 sized in practice, so the 0 value would do fine to determine if set or not.

incus-osd/api/system_kernel.go
```diff
@@ -18,8 +18,8 @@ type SystemKernelConfigNetwork struct {
 	BufferSize             int    `json:"buffer_size,omitempty"              yaml:"buffer_size,omitempty"`
 	QueuingDiscipline      string `json:"queuing_discipline,omitempty"       yaml:"queuing_discipline,omitempty"`
 	TCPCongestionAlgorithm string `json:"tcp_congestion_algorithm,omitempty" yaml:"tcp_congestion_algorithm,omitempty"`
-	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty" yaml:"netdev_max_backlog,omitempty"`
-	TCPMTUProbing          bool   `json:"tcp_mtu_probing,omitempty" yaml:"tcp_mtu_probing,omitempty"`
+	NetdevMaxBacklog       int    `json:"netdev_max_backlog,omitempty"       yaml:"netdev_max_backlog,omitempty"`
+	TCPMTUProbing          bool   `json:"tcp_mtu_probing,omitempty"          yaml:"tcp_mtu_probing,omitempty"`
```
stgraber, line 22: You're fixing the indent in this commit, those indent fixes should instead be done in their respective earlier commits.

stgraber:
This PR adds the API fields for those flags but it doesn't implement any of the actual setting logic. Marking PR as draft for now.

sharathsivakumar:
Apologies, I intended it to be a draft.  
Ah. I am missing something here. Let me get back to you on this.

sharathsivakumar:
@stgraber I have made the changes that implement netdev_max_backlog and tcp_mtu_probing. Please let me what you think about those. 
A few comments/Questions:
1. With regard to the commits and order, please ignore them for now. I will use this PR as a draft for review and clean it up once I am sure about all the changes. 
2.  I have not added any code that validates `tcp_mtu_probing`. In the sense that if the value is set to `truee` or anything with a typo, currently, the value will not be set. But since the default is 0 and the current implementation sets it to 1 only when its `true` , it falls on the user to figure out why the value was not being set. Is this fine?
3. With regard to setting ring buffer sizes, I can take the same approach of setting the values based on the yaml config. The requester has also mentioned about setting a max value by auto-discovering the max value. I am still new to this and I will check with the requester on how the value can be obtained. Alternatively, please feel free to point me to documentation that does this. Also, if auto-discover is not feasible, I can proceed with the normal approach of setting the values.
PS: I have requested for more information from the requester as well for the ring buffer sizing

sharathsivakumar:
Will be opening a new draft PR

PR #1049, opened 2026-04-18
Add info subcommands for system endpoints

### LINKED ISSUE #504: Implement some `info` commands in the CLI
Right now we have the usual `show` and `edit` which directly show and push the API values.

This isn't necessarily the most user friendly approach as the API may contain a bunch of extra data which doesn't really make sense to expose to the user. Also the type of some of the data, especially thing like the `size` fields can be hard to read as those are usually raw byte counts.

An `info` command can then return a more human readable version of the data, including showing tables, using colors, ... `info` would be deemed unsuitable for parsing/scripting as `show` should be used instead for that.

pranav767:
Resolves: https://github.com/lxc/incus-os/issues/504

This PR adds info sub commands to system endpoint(s).
As mentioned in the Issue, I've extended `subCommand` to effectively add an `info` member as a function to be run for the `info` command.
- A New file `cli_system_info.go` is created which has implementation of all info subcommands (logging,network, provider,resources, security, storage & update)
- A common helper function which acts as a boilerplate for all info commands, i.e. this function builds the `apiURL` and makes a http request to the api endpoint, and deserializes the response.
- Other functions are just to format and view the required values properly. YAML dump (logging, provider, security, update)

There might be some deviations from the below mentioned `info.go` such as breaking info more functions (purely done to avoid failures with `make static-analysis`
For reference:

`system resources info` looks exactly like ` incus info --resources` from 
https://github.com/lxc/incus/blob/main/cmd/incus/info.go#L382

### DIFF
--- a/incus-osd/.golangci.yml
+++ b/incus-osd/.golangci.yml
@@ -29,6 +29,7 @@ linters:
 
     gosec:
       excludes:
+      - G115
       - G304
       - G701
       - G702
--- a/incus-osd/cli/cli_system.go
+++ b/incus-osd/cli/cli_system.go
@@ -94,6 +94,7 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 		description   string
 		isWritable    bool
 		extraCommands func() []*cobra.Command
+		info          func(os *cmdAdminOS, endpoint string, description string) *cobra.Command
 	}
 
 	subCommands := []subCommand{
@@ -106,11 +107,13 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "logging",
 			description: "System logging",
 			isWritable:  true,
+			info:        systemInfoCommand,
 		},
 		{
 			name:        "network",
 			description: "Network configuration",
 			isWritable:  true,
+			info:        systemInfoNetworkCommand,
 			extraCommands: func() []*cobra.Command {
 				// Confirm new network configuration.
 				networkConfirmCmd := cmdGenericRun{
@@ -135,16 +138,19 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "provider",
 			description: "Image and management provider",
 			isWritable:  true,
+			info:        systemInfoCommand,
 		},
 		{
 			name:        "resources",
 			description: "System resources",
 			isWritable:  false,
+			info:        systemInfoResourcesCommand,
 		},
 		{
 			name:        "security",
 			description: "Security configuration",
 			isWritable:  true,
+			info:        systemInfoCommand,
 			extraCommands: func() []*cobra.Command {
 				// TPM rebind.
 				tpmRebindCmd := cmdGenericRun{
@@ -162,6 +168,7 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "storage",
 			description: "Storage configuration",
 			isWritable:  true,
+			info:        systemInfoStorageCommand,
 			extraCommands: func() []*cobra.Command {
 				// Create storage volume.
 				createVolumeCmd := cmdGenericRun{
@@ -250,6 +257,7 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "update",
 			description: "Update configuration",
 			isWritable:  true,
+			info:        systemInfoCommand,
 			extraCommands: func() []*cobra.Command {
 				// Check updates.
 				checkUpdatesCmd := cmdGenericRun{
@@ -282,6 +290,11 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 		// Show.
 		showCmd := cmdGenericShow{os: c.os, endpoint: "system/" + sub.name}
 		subCmd.AddCommand(showCmd.command())
+
+		// Info.
+		if sub.info != nil {
+			subCmd.AddCommand(sub.info(c.os, "system/"+sub.name, "Show "+sub.description))
+		}
 
 		cmd.AddCommand(subCmd)
 
--- a/incus-osd/cli/cli_system.go
+++ b/incus-osd/cli/cli_system.go
@@ -108,6 +108,7 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "kernel",
 			description: "System kernel configuration",
 			isWritable:  true,
+			info:        systemInfoKernelCommand,
 		},
 		{
 			name:        "logging",
--- a/incus-osd/cli/cli_system.go
+++ b/incus-osd/cli/cli_system.go
@@ -10,6 +10,7 @@ type cmdAdminOSSystem struct {
 	os *cmdAdminOS
 }
 
+//nolint:revive
 func (c *cmdAdminOSSystem) command() *cobra.Command {
 	cmd := &cobra.Command{}
 	cmd.Use = cli.Usage("system")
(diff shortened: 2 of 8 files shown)


incus-osd/.golangci.yml
```diff
@@ -29,6 +29,7 @@ linters:
 
     gosec:
       excludes:
+      - G115
```
stgraber: I think I'd prefer we put `// #nosec G115` where needed rather than turn it off entirely.

incus-osd/cli/cli_system_info.go
```diff
@@ -0,0 +1,703 @@
...
+			return err
+		}
+
+		_, _ = fmt.Printf("%s", out) //nolint:forbidigo
+
+		return nil
+	})
+}
+
+// systemInfoNetworkCommand returns an info command for the system/network endpoint.
```
stgraber: I think it would make sense to move each info command to its own file, it will help a fair bit with readability when it comes to the resources one at least :)

incus-osd/cli/cli_system.go
```diff
@@ -106,11 +107,13 @@ func (c *cmdAdminOSSystem) command() *cobra.Command {
 			name:        "logging",
 			description: "System logging",
 			isWritable:  true,
+			info:        systemInfoCommand,
```
stgraber: Is there any actual difference between `incus admin os system logging show` and `incus admin os system logging info` at this stage?

If not, I'd argue that we just shouldn't include an `info` command where it doesn't bring in any extra value.

incus-osd/cli/cli_system_info.go
```diff
@@ -0,0 +1,703 @@
...
+	cli "github.com/lxc/incus/v6/shared/cmd"
+	"github.com/lxc/incus/v6/shared/units"
+	"github.com/spf13/cobra"
+	"go.yaml.in/yaml/v4"
+
+	"github.com/lxc/incus-os/incus-osd/api"
+)
+
+// makeInfoCommand is a generic helper that handles the common boilerplate for info commands.
+func makeInfoCommand[T any](c *cmdAdminOS, endpoint string, description string, render func(T) error) *cobra.Command {
```
stgraber: That function should probably go in the same file as the other generic wrappers.

incus-osd/cli/cli_system_info.go
```diff
@@ -0,0 +1,703 @@
...
+		}
+
+		return render(data)
+	}
+
+	return cmd
+}
+
+// systemInfoCommand returns an info command that renders the endpoint as YAML.
+func systemInfoCommand(c *cmdAdminOS, endpoint string, description string) *cobra.Command {
```
stgraber: This one can probably go away assuming that the `info` commands using it where otherwise identical to the `show` command.

stgraber:
I haven't directly tested the code myself but this looks good to me.

Main changes I think should be:
 - Create one separate file for each of the three initial `info` commands (network, resources and storage)
 - If the other `info` commands are identical to the current `show` commands, we shouldn't include them at all
 - Move the generic logic into the same file as the other generic wrappers
 - Selectively nolint G115 rather than disabling it for the whole repo

I have a feeling we could re-shuffle things a little to have the new Info logic better align with the existing cmdGenericXYZ wrappers, but I don't know how much work would be involved in doing that, so the current approach is fine to keep too.

I think I'd personally have done a new `cmdGenericinfo` which would take a near identical set of fields as `cmdGenericShow` except for an extra field that would hold a function doing the actual custom Info handling. The generic layer would then do:
 - Command definition
 - Basic argument parsing and validation
 - GET API call
 - Then call the function with that data so the function can take care of the custom data processing and rendering

I think that would result in a slightly smaller change than is currently present in this PR and would be more in line with the current generic function approach.

incus-osd/cli/cli_system.go
pranav767: Nope there is no actual difference, Yes I'll remove similar commands.

pranav767:
Apologies for the delayed response :bow: , I'll fix the changes with suggested way & follow ` existing cmdGenericXYZ wrappers`

> Main changes I think should be:
> 
>     * Create one separate file for each of the three initial `info` commands (network, resources and storage)
> 
>     * If the other `info` commands are identical to the current `show` commands, we shouldn't include them at all
> 
>     * Move the generic logic into the same file as the other generic wrappers
> 
>     * Selectively nolint G115 rather than disabling it for the whole repo

pranav767:
@stgraber I've updated the PR, with recommended changes to have a `cmdGenericinfo` instead. Also separate files for `network, storage and resources`, dropping others as those are similar to `show`
Testing `info` command:

```
jinx@jinx:~/opensource/incus$ ./incus admin os system storage info
WARNING: The IncusOS API and configuration is subject to change

Drives:
+-----------------------------------------------------+---------------+------------+------+----------+------+-----------+-------------+
|                         ID                          |     MODEL     |   SERIAL   | BUS  | CAPACITY | BOOT | ENCRYPTED | MEMBER POOL |
+-----------------------------------------------------+---------------+------------+------+----------+------+-----------+-------------+
| /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root | QEMU HARDDISK | incus_root | scsi | 50.00GiB | yes  |           | local       |
+-----------------------------------------------------+---------------+------------+------+----------+------+-----------+-------------+

Pools:
+-------+-----------+--------+----------------+----------+-------------+-----------+------------------------------------------------------------+
| NAME  |   TYPE    | STATE  | ENCRYPTION KEY | RAW SIZE | USABLE SIZE | ALLOCATED |                          DEVICES                           |
+-------+-----------+--------+----------------+----------+-------------+-----------+------------------------------------------------------------+
| local | zfs-raid0 | ONLINE | available      | 16.50GiB | 16.50GiB    | 4.35MiB   | /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_root-part11 |
+-------+-----------+--------+----------------+----------+-------------+-----------+------------------------------------------------------------+
jinx@jinx:~/opensource/incus$ ./incus admin os system network info
WARNING: The IncusOS API and configuration is subject to change

+--------+-----------+----------+-------------------+------+-------+-----------------------------------------+------------+
|  NAME  |   TYPE    |  STATE   |      HWADDR       | MTU  | SPEED |                ADDRESSES                |   ROLES    |
+--------+-----------+----------+-------------------+------+-------+-----------------------------------------+------------+
| enp5s0 | interface | routable | 10:66:6a:80:be:4e | 1500 | -1    | 10.63.169.35                            | management |
|        |           |          |                   |      |       | fd42:e6bf:9105:f144:1266:6aff:fe80:be4e | cluster    |
+--------+-----------+----------+-------------------+------+-------+-----------------------------------------+------------+

./incus admin os system resources info 
WARNING: The IncusOS API and configuration is subject to change

System:
  UUID: ae4be395-caad-4809-ac0c-43d2ea7084ed
  Vendor: QEMU
  Product: Standard PC (Q35 + ICH9, 2009)
  Version: pc-q35-8.2
  Type: virtual-machine
  Chassis:
      Vendor: QEMU
      Type: Other
      Version: pc-q35-8.2
  Motherboard:
      Vendor: LinuxContainers
      Product: Incus
      Version: pc-q35-8.2
  Firmware:
      Vendor: Ubuntu distribution of EDK II
      Version: 2024.02-2ubuntu0.7
      Date: 11/27/2025

Load:
  Processes: 192
  Average: 0.13 0.04 0.01

CPU:
  Architecture: x86_64
  Vendor: GenuineIntel
...
```

Again apologies for the delay :bow:

pranav767:
Hey @stgraber, 
could you please review this when you have some time, don't think there are any more expected changes.

stgraber:
Sorry for the delay, looks great!

PR #1058, opened 2026-04-23
Introduce application-specific config structs

gibmat:
Refactor our handling of application state from a generic map of applications to application-specific structs.


incus-osd/internal/applications/sysext.go
```diff
@@ -0,0 +1,295 @@
+package applications
```
stgraber, line 1: That part looks like it's in the wrong commit?

Issue #1063, opened 2026-04-25
API/command to display the Tailscale node key


strawberry192:
Tailscale has a security feature called [tailnet lock](https://tailscale.com/docs/features/tailnet-lock), which requires new nodes to be signed by a trusted signing node before connections can be made, and offers some protection against Tailscale's backend or accounts being compromised. 

However, to sign the node key securely we must be able to retrieve it locally, as otherwise Tailscale's server can, in principle, provide a malicious one to MITM the node, negating any security benefit.

On a typical linux server the nodekey can be obtained by: `tailscale status --json | jq .Self.PublicKey`. This is a feature request to provide an API or CLI command to retrieve it (it would be enough to include it in the tailscale state/config).

stgraber:
IncusOS services can return state information. Returning a subset of `status --json` as the service state would make sense I think.

I'd probably want to see something like:
```
type ServiceTailscaleStatePeer {
    IPAddresses []string // TailscapeIPs
    FQDN string // DNSName
    Online bool // Online
    PublicKey string // PublicKey
    ExitNode bool // ExitNode
    RxBytes int64 // RxBytes
    TxBytes int64 // TxBytes
}

type ServiceTailscaleState struct {
    Local ServiceTailscaleStatePeer // Self
    Tailnet string // CurrentTailnet.Name
    TailnetDNS string // CurrentTailnet.MagicDNSSuffix
    Peers []ServiceTailscaleStatePeer // Peer
}
```

That would surface the most relevant state information which isn't already covered in configuration.

strawberry192:
Seems to work now, thanks!

PR #1069, opened 2026-04-28
incus-osd/services: Add Tailscale state Support

believerd:
New packages are introduced by `tailscale.com`. It also bumps the golang version. I don't know if this is a concern?

Does it make sense to be able to "edit" state info when using `incus admin os service edit` command?
Maybe we should hide the state information in `edit` command.

#1063


stgraber:
> Does it make sense to be able to "edit" state info when using `incus admin os service edit` command?

The state is always read-only but it still shown as it can be useful to see when modifying the writable config section.

incus-osd/api/service_tailscale.go
```diff
@@ -17,3 +21,42 @@ type ServiceTailscaleConfig struct {
...
+// ServiceTailscaleState represents the current state of the Tailscale service.
+type ServiceTailscaleState struct {
+	Version          string                           `json:"version"`
+	BackendState     ServiceTailscaleBackendStateEnum `json:"backend_state"`
+	TailnetName      string                           `json:"tailnet_name,omitempty"`
+	TailnetDNSSuffix string                           `json:"tailnet_dns_suffix,omitempty"`
+	Self             ServiceTailscaleStatePeer        `json:"self"`
+	Peers            []ServiceTailscaleStatePeer      `json:"peer,omitempty"`
+	HaveNodeKey      bool                             `json:"have_node_key"`
+	Health           []string                         `json:"health,omitempty"`
```
stgraber, line 61: What are these two?

incus-osd/internal/util/tailscale.go
```diff
@@ -0,0 +1,84 @@
+package util
```
stgraber: All that code is only ever called from the Tailscale service Get function in the tailscale service. So I'd expect all of it to be moved to within the services package with those functions moved as un-exported functions on the Tailscale struct so we keep things nicely namespaced.

incus-osd/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/lxc/incus-os/incus-osd
```
stgraber, line 1: gomod (go.mod+go.sum) updates are done as their own commit.

stgraber:
Looks good, just a couple questions and a bit of code re-shuffling needed.

incus-osd/api/service_tailscale.go
believerd: Here's an example of `health` in `service show`. They are some useful information from Tailscale.
```
   health:
        - Some peers are advertising routes but --accept-routes is false
```

As for `HaveNodeKey`, I think it's a flag indicating whether there's a key pair for connecting to tailnet which I assume will always be true, but there might be some cases I don't know yet so I just leave it there for now.

I get those names from Tailscale's struct directly but it might make sense to link the struct file for reference.
https://github.com/tailscale/tailscale/blob/main/ipn/ipnstate/ipnstate.go

incus-osd/internal/util/tailscale.go
believerd: Got you. I was thinking about keeping those logic out of service package to keep it clean.

believerd:
> > Does it make sense to be able to "edit" state info when using `incus admin os service edit` command?
> 
> The state is always read-only but it still shown as it can be useful to see when modifying the writable config section.

Yeah, I know it won't actually do anything but just the idea that it can be modified when in the editing view makes it feel a bit contradictory to me. It might be useful if we could have some comments in the editing view saying that it won't do anything to the state changes. Just a personal preference ig.

PR #1073, opened 2026-04-29
Preparation for separate HypervisorOS branding

gibmat:
A variety of changes to remove hard-coded occurrences of "IncusOS" and other preparatory cleanup work to make switching to HypervisorOS branding easier.


mkosi.packages/initrd-utils/debian/control
```diff
@@ -14,6 +14,6 @@ Depends: kpartx,
          swtpm-tools,
          ${misc:Depends},
          ${shlibs:Depends}
-Description: IncusOS initrd utilities
- Various utilities, services, and configuration files needed in the IncusOS
- initrd.
+Description: initrd utilities for IncusOS/HypervisorOS
+ Various utilities, services, and configuration files needed in the IncusOS/
+ HypervisorOS initrd.
```
stgraber: I don't think that we ever want to see the HypervisorOS string in the IncusOS repository. Given the initrd-utils package isn't really user visible, its description can include IncusOS without needing any rewrite logic I think.

Makefile
```diff
@@ -248,6 +248,11 @@ doc-setup:
 	rm -Rf doc/html
 	rm -Rf doc/.sphinx/.doctrees
 
+	find ./doc/ -name '*.md' | xargs sed -i -e "s@github.com/lxc/incus-os/blob/main/@github.com/lxc/incus-os/blob/main/@g"
+	find ./doc/ -name '*.md' | xargs sed -i -e "s/IncusOS/IncusOS/g"
+	find ./doc/ -name '*.md' | xargs sed -i -e "s/INCUSOS/INCUSOS/g"
+	find ./doc/ -name '*.md' | xargs sed -i -e "s/incusos/incusos/g"
```
stgraber: I don't think we should rewrite the github URL as HypervisorOS is a private repository so nobody who reads the doc will be able to follow the link anyway :)

For the others, it would be nice if we could derive them from OSNAME.
Assuming we have OSNAME as IncusOS, we can derive the first as an all-caps version of it and the latter as an all-lower version of it.

incus-osd/cmd/incus-osd/main.go
```diff
@@ -49,44 +49,44 @@ func main() {
 
 	// Check privileges.
 	if os.Getuid() != 0 {
-		tui.EarlyError("incus-osd must be run as root")
+		tui.EarlyError("incus-osd must be run as root", "IncusOS")
```
stgraber: Can we grab the OS name earlier so we don't have to hardcode+sed all of those?

incus-osd/internal/providers/provider_local.go
```diff
@@ -94,7 +101,7 @@ func (p *local) GetOSUpdate(ctx context.Context) (OSUpdate, error) {
 	foundUpdateFile := false
 
 	for _, asset := range p.releaseAssets {
-		if strings.HasPrefix(filepath.Base(asset), "IncusOS_") && strings.Contains(filepath.Base(asset), p.releaseVersion) {
+		if strings.HasPrefix(filepath.Base(asset), osName+"_") && strings.Contains(filepath.Base(asset), p.releaseVersion) {
```
stgraber: Don't we have access to the state struct from this spot (p.state or something like that)?

incus-osd/internal/install/install.go
```diff
@@ -1174,8 +1174,14 @@ func configureSWTPM(ctx context.Context, isInstall bool) error {
 
 // initializeSWTPM initializes the swtpm state in the given root directory.
 func initializeSWTPM(ctx context.Context, swtpmRoot string) error {
```
stgraber, line 1176: Wouldn't it be cleaner to pipe through the state struct so we can fetch the OS name from it?

It looks like we'd need to pass it to CheckSystemRequirements which would pass it to configureSWTPM and finally initializeSWTPM

incus-osd/internal/providers/provider_local.go
gibmat: Duh! Yes, we've got access to state. :)

incus-osd/internal/install/install.go
gibmat: I'd argue it's cleaner not to pass through a pointer through multiple levels of calls, and instead directly get the information needed via `GetCurrentRelease()`. Somewhat astonishingly we don't have `state.State` intertwined in the `install` package, and I don't like adding it just for this, especially when fetching the release information is very quick.

stgraber: Fair enough

stgraber (approved):

PR #1074, opened 2026-05-01
image-customizer: Add support for updates and rescue media

masnax:
* Adds an auto-expiring map for prepared update caches (defaults to 10 mins per entry)

* `POST /1.0/images` can now return a second `resources` path if `offline=true` in the request data.
  - The content of the generated image will be the set of application names in the seed data.
  - Can now also filter by version in addition to architecture and channel

* `POST /1.0/updates` returns a tarball containing all resources for a specific channel + version. Can be further filtered by providing an allow-list of architectures, components, and component types.

`GET /1.0/images/{uuid}` has been renamed to `GET /1.0/files/{uuid}`

### DIFF
--- a/incus-osd/api/customizer/images.go
+++ b/incus-osd/api/customizer/images.go
@@ -1,15 +1,19 @@
 package customizer
 
 import (
+	apiimages "github.com/lxc/incus-os/incus-osd/api/images"
 	apiseed "github.com/lxc/incus-os/incus-osd/api/seed"
 )
 
 // ImagesPost represents the data needed for POST /1.0/images.
 type ImagesPost struct {
-	Architecture string          `json:"architecture" yaml:"architecture"`
-	Type         string          `json:"type"         yaml:"type"`
-	Seeds        ImagesPostSeeds `json:"seeds"        yaml:"seeds"`
-	Channel      string          `json:"channel"      yaml:"channel"`
+	Architecture apiimages.UpdateFileArchitecture `json:"architecture" yaml:"architecture"`
+	Type         string                           `json:"type"         yaml:"type"`
+	Channel      string                           `json:"channel"      yaml:"channel"`
+	Version      string                           `json:"version"      yaml:"version"`
+
+	Seeds   ImagesPostSeeds `json:"seeds"   yaml:"seeds"`
+	Offline bool            `json:"offline" yaml:"offline"`
 }
 
 // ImagesPostSeeds represents the seed data included in a ImagesPost request.
@@ -22,4 +26,19 @@ type ImagesPostSeeds struct {
 	OperationsCenter *apiseed.OperationsCenter `json:"operations-center" yaml:"operations-center"` //nolint:tagliatelle
 	Provider         *apiseed.Provider         `json:"provider"          yaml:"provider"`
 	Update           *apiseed.Update           `json:"update"            yaml:"update"`
+}
+
+// UpdatesPost represents the data needed for POST /1.0/updates.
+type UpdatesPost struct {
+	UpdateFilter `yaml:",inline"`
+}
+
+// UpdateFilter represents filterable parameters for update resources.
+type UpdateFilter struct {
+	Channel string `json:"channel" yaml:"channel"`
+	Version string `json:"version" yaml:"version"`
+
+	Components    []apiimages.UpdateFileComponent    `json:"components"    yaml:"components"`
+	Types         []apiimages.UpdateFileType         `json:"types"         yaml:"types"`
+	Architectures []apiimages.UpdateFileArchitecture `json:"architectures" yaml:"architectures"`
 }
--- a/incus-osd/cmd/image-customizer/main.go
+++ b/incus-osd/cmd/image-customizer/main.go
@@ -19,10 +19,10 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/lxc/incus/v6/shared/subprocess"
 	"github.com/pires/go-proxyproto"
 	"github.com/timpalpant/gzran"
 	"go.yaml.in/yaml/v4"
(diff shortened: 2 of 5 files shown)


incus-osd/api/customizer/images.go
```diff
@@ -1,15 +1,24 @@
...
 // ImagesPost represents the data needed for POST /1.0/images.
 type ImagesPost struct {
-	Architecture string          `json:"architecture" yaml:"architecture"`
-	Type         string          `json:"type"         yaml:"type"`
-	Seeds        ImagesPostSeeds `json:"seeds"        yaml:"seeds"`
-	Channel      string          `json:"channel"      yaml:"channel"`
+	IndexOptions `yaml:",inline"`
+
+	Seeds   ImagesPostSeeds `json:"seeds"   yaml:"seeds"`
+	Bundled bool            `json:"bundled" yaml:"bundled"`
```
stgraber: What's that one about?

incus-osd/api/customizer/images.go
```diff
@@ -23,3 +32,25 @@ type ImagesPostSeeds struct {
...
+
+// ImageRescuePost represents the data needed for POST /1.0/images?type=rescue.
+type ImageRescuePost struct {
+	IndexOptions `yaml:",inline"`
+
+	Applications []apiimages.UpdateFileComponent `json:"applications" yaml:"applications"`
+}
+
+// UpdatesPost represents the data needed for POST /1.0/images?type=update.
+type UpdatesPost struct {
```
stgraber, line 32: Naming suggests it would be at /1.0/updates which would probably make more sense anyway. What's the reason for pushing everything through /1.0/images?

incus-osd/api/customizer/images.go
```diff
@@ -23,3 +32,25 @@ type ImagesPostSeeds struct {
 	Provider         *apiseed.Provider         `json:"provider"          yaml:"provider"`
 	Update           *apiseed.Update           `json:"update"            yaml:"update"`
 }
+
+// ImageRescuePost represents the data needed for POST /1.0/images?type=rescue.
```
stgraber: For this one, what do you think of having POST /1.0/images return multiple URLs?

That way the request for both the primary install media and the secondary application media would be processed together, allowing for a simple Offline boolean flag in the request to tell us whether to generate an additional image and what to put in there (as the main request already tells us the type of image, architecture, channel and application list).

incus-osd/api/customizer/images.go
masnax: Oops, remnant from an earlier approach :)

incus-osd/api/customizer/images.go
masnax: Yeah, we can do that. I suppose the application list would just be ignored if `offline=false`.

incus-osd/api/customizer/images.go
masnax: The initial `POST` doesn't care at all about the output, it just caches the supplied options against a UUID, so it doesn't make much sense creating separate API paths that perform the same thing. 

The subsequent `GET` reads from the cached data, which is where the context about which file should be returned lives, so it's also not dependent on any context from the `GET` request beyond the UUID.

So in essence, the two API actions feel more like "POST file request" and "GET file by request UUID". I can split them up by file type if you'd prefer, though it would be more duplicated code. I can also move the type parameter into the request data instead of using a query param.

incus-osd/api/customizer/images.go
stgraber: Ah no, I meant we already have the application list anyway in the form of the applications seed. So we don't even need an additional field for it.

stgraber:
@masnax from an API point of view, I think it'd be cleaner to have distinct:
 - /1.0/images
 - /1.0/updates

Each with their own distinct request struct.

The images one is what we have today but with the addition of the `Offline` field where when it's set to `true` will have us provide two URLs back rather than just the one. As mentioned above, we don't need any other data from the user as the normal image request already gives us the type of media, architecture, channel and the list of applications that will be installed on boot.

The updates one gets a new struct which lets you provide an image identifier in the form of either a specific version or a channel name and then a filter so you can fetch just a subset of the update tree.

masnax:
OK, I've split the `POST` into two parts:

* `POST /1.0/images` returns 1 or 2 URLs keyed by `image` and `resources` respectively. 
* `POST /1.0/updates` returns 1 URL keyed by `update`

I renamed `GET /1.0/images/{uuid}` to `GET /1.0/files/{uuid}` to be a bit more general (since updates don't return images). Doesn't seem this affects the UI for the customizer since the logic just tacks the `image` path to the URL.

incus-osd/api/customizer/images.go
```diff
@@ -23,3 +27,18 @@ type ImagesPostSeeds struct {
...
+// UpdatesPost represents the data needed for POST /1.0/updates.
+type UpdatesPost struct {
+	Filter UpdateFilter `json:"filter" yaml:"filter"`
+}
+
+// UpdateFilter represents filterable parameters for update resources.
+type UpdateFilter struct {
+	Channel string `json:"channel" yaml:"channel"`
+	Version string `json:"version" yaml:"version"`
+
```
stgraber, line 40: I think I'd move those to UpdatesPost itself as they are used to select the image itself whereas the rest of the fields are the actual filters being applied to the image's content.

incus-osd/cmd/image-customizer/main.go
```diff
@@ -343,97 +359,106 @@ func apiImage(w http.ResponseWriter, r *http.Request) {
 	// Image UUID.
 	imageUUID := r.PathValue("uuid")
 
-	imagesMu.Lock()
+	opts, ok := images.Get(imageUUID)
```
stgraber: I guess `images` here should become `files` for consistency

stgraber:
@masnax can you also take this change in here to get the doc tests passing?
https://github.com/lxc/incus/pull/3285/changes/b56b7fc6be35e9148539f8e4ce54d47afc87411f

incus-osd/api/customizer/images.go
masnax: That struct is used in all 3 cases to filter the returned assets. I'll make it embedded.

PR #1078, opened 2026-05-05
Disable incus-agent exec and file transfers by default

### LINKED ISSUE #990: Disable incus-agent exec and file transfer support by default, re-enable through systemd-cred
By default the Incus agent should be configured to disable exec and file transfers, then a systemd credential would allow re-enabling that on the VM but will have it flagged in a similar way to when booted with Secure Boot or TPM 2.0 disabled.

gibmat:
By default, configure incus-agent to disable exec and file transfers, but retain other capabilities. All features can be enabled by setting the `fully-enable-incus-agent=true` systemd-credential in the VM's configuration, but that will trigger IncusOS to report a degraded security state.

Closes #990

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -132,6 +132,7 @@ jobs:
             -c limits.memory=2GiB \
             -d root,size=50GiB
           incus config device add test-incus-os vtpm tpm
+          incus config set test-incus-os systemd.credential.fully-enable-incus-agent true
           incus start test-incus-os
 
           incus wait test-incus-os agent --timeout 300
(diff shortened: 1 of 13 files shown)


stgraber:
Couple questions here:
1) Isn't there a time window here where the agent will have exec capabilities during boot which could allow for a race where the user can get an exec or file session prior to the enforcement kicking in?
2) Is `systemctl restart incus-agent` safe to do on a non-Incus system? We may want to make the whole thing conditional on the agent virtio-ports entry existing.

gibmat:
> 1. Isn't there a time window here where the agent will have exec capabilities during boot which could allow for a race where the user can get an exec or file session prior to the enforcement kicking in?

Yes, there would have been a very brief window on first boot; I've added configuration for `tmpfiles.d` to automatically install a default incus-agent configuration during early first boot before the agent would have a chance to start.

> 2. Is `systemctl restart incus-agent` safe to do on a non-Incus system? We may want to make the whole thing conditional on the agent virtio-ports entry existing.

Good point, I've gated the entire function behind that file existing in `/dev/`.

PR #1080, opened 2026-05-07
Provide incus client versions in requirements.md

goetzk:
`incus admin os` is not available in all versions of the incus client but is needed for some management tasks, such as retrieving volume encryption keys. This change adds a note to the requirements page so administrators know to upgrade their clients.

I'm not sure if githubs signed-off-by in the commit is hiding my email or didn't include it at all; i'm happy to pull the repo and re sign off if its a problem.

### DIFF
--- a/doc/getting-started/requirements.md
+++ b/doc/getting-started/requirements.md
@@ -35,3 +35,13 @@ Certain physical servers and Microsoft Hyper-V have known incomplete and/or brok
 ### Using a software-backed TPM
 
 Most consumer-grade ARM systems lack physical TPM chips, and it can be cost prohibitive to purchase one, such as when using a RaspberryPi. To support these systems, IncusOS can utilize `swtpm` for its TPM. For further details, see [Installing without a TPM](../reference/installing-without-tpm.md).
+
+## Compatible clients
+
+Some incus-os administrative functions can be accessed via `incus admin os ...`. `incus admin os` was added to Incus in the following client versions:
+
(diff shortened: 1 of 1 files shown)


stgraber:
I'm not sure that it really makes sense to list those versions given that all supported versions of Incus includes the `incus admin os` command and we don't support folks running unsupported versions.

I'd have felt differently before this was backported and released within the 6.0 LTS builds, but currently the only two supported versions of Incus are 6.0.6 and 7.0, both of which support it. Even if we expand it to the most recently out of support release, 6.23, it's also supported on there.

stgraber:
It's also a bit misleading to add the `incus` client as a requirement when IncusOS deployments with either Operations Center or Migration Manager as the primary application will instead use the `operations-center` or `migration-manager` command line tool instead, both of which also have the `admin os` command.

Lastly, you can also manage IncusOS without the CLI by using the OS page within the Incus UI.

goetzk:
> I'm not sure that it really makes sense to list those versions given that all supported versions of Incus includes the `incus admin os` command and we don't support folks running unsupported versions.
> 
> I'd have felt differently before this was backported and released within the 6.0 LTS builds, but currently the only two supported versions of Incus are 6.0.6 and 7.0, both of which support it. Even if we expand it to the most recently out of support release, 6.23, it's also supported on there.

I'm not super attached to this change so won't worry about this further - especially since the incus client has all the bits in it for ~6 months.

thanks for your quick response .

PR #1081, opened 2026-05-07
Further cleanup in preparation for HypervisorOS

gibmat:


.github/workflows/build.yml
```diff
@@ -101,8 +101,24 @@ jobs:
...
           path: upload/
+
+      - name: Upload binaries to local server
+        env:
+          UPLOAD_HOST: ${{ secrets.UPLOAD_HOST }}
+        if: ${{ env.UPLOAD_HOST != '' }}
+        run: |
+          mkdir ~/.ssh/
+          chmod 700 ~/.ssh/
+          echo "${{ secrets.UPLOAD_HOST }} ${{ secrets.UPLOAD_HOST_CERT }}" > ~/.ssh/known_hosts
```
gibmat: I don't know if this will be quite right for pre-populating the `known_hosts` file

.github/workflows/build.yml
```diff
@@ -101,8 +101,24 @@ jobs:
...
+          echo "${{ secrets.UPLOAD_CLIENT_KEY }}" > ~/.ssh/id_ed25519.pub
+          chmod 600 ~/.ssh/id_ed25519.pub
+
+          scp upload/* ${{ secrets.UPLOAD_HOST }}:${{ secrets.UPLOAD_PATH }}
```
gibmat: Not sure what username should be used when connecting to the upload host?

.github/workflows/build.yml
stgraber: It won't be as our UPLOAD_HOST contains square brackets.

I think easiest is to just write UPLOAD_HOST_CERT into the file and we'll adjust that env variable to have the correct format for the known_hosts file

.github/workflows/build.yml
stgraber: Oh, good point, I'll add a secrets.UPLOAD_USERNAME key as well

stgraber:
@gibmat I've updated the secrets so that UPLOAD_USERNAME is now set and UPLOAD_HOST_CERT contains the entry as it should be in the known_hosts file

.github/workflows/build.yml
```diff
@@ -101,8 +101,24 @@ jobs:
...
+      - name: Upload binaries to local server
+        env:
+          UPLOAD_HOST: ${{ secrets.UPLOAD_HOST }}
+        if: ${{ env.UPLOAD_HOST != '' }}
+        run: |
+          mkdir ~/.ssh/
+          chmod 700 ~/.ssh/
+          echo "${{ secrets.UPLOAD_HOST_CERT }}" > ~/.ssh/known_hosts
+          echo "${{ secrets.UPLOAD_CLIENT_KEY }}" > ~/.ssh/id_ed25519.pub
+          chmod 600 ~/.ssh/id_ed25519.pub
```
stgraber: That looks wrong, the key you have in secret is the private key, so shouldn't have the pub suffix

gibmat: Doh!

stgraber (approved):

PR #1090, opened 2026-05-12
doc: Separate download and support pages into individual IncusOS/HypervisorOS files

gibmat:
Sphinx has an `only` directive that can conditionally include content when rendering documents. However, it only handles plain text sections; anything with section headers or other blocks of text either don't render or render incorrectly.

This PR divides up the download and support pages, with symlinks for for IncusOS. On the HypervisorOS side, we can introduce equivalent pages and just change the symlinks.


stgraber:
@gibmat any reason we need to have the hypervisor os files in this repo?

Maybe I missed something but it looks like we've just split the branding-specific content to a separate _BRAND page which we then include from the generic page.

So we shouldn't need to have all the brand versions in the main repo and can instead just include them in their respective repo with the only delta being the change of "include" statement.

Though at that point I'm wondering if we couldn't do something even simpler.
Basically have `download_incusos.md` and `download.md` being a symlink to `download_incusos.md`.

We'd then have the brand repo just introduce `download_hypervisoros.md` and change the symlink, avoiding any of the potential text conflicts.

stgraber:
@gibmat can you update this one?

PR #1097, opened 2026-05-14
Report a friendly version of each application

### LINKED ISSUE #1094: Add support for applications exposing the software version
Currently the application version is its image version (YYYYMMDDhhmm).
That's not particularly user friendly as it doesn't line up with the version of the software that they're used to seeing in release announcements.

The manifest allows for tracking things down and figuring it out. But that's not always easy to get your hands on or to read.

It would be good for the application interface to have a function returning the software version.
Calling this on Incus would get you 7.0.0 for example. It should be possible to implement such logic on most of our current applications. Those where it's not possible should just return an empty string.

This should then be exposed in the application API alongside the primary version field (image version).

gibmat:
This adds the ability for applications to report a "friendly" version as a free form string. For Incus, it might look something like "7.0.0 [202605191531]", reporting the actual application version and the build version in brackets. This field isn't intended to be parsed, and its format may change at any time.

image

Closes #1094

### DIFF
--- a/app-build/applications.json
+++ b/app-build/applications.json
@@ -121,7 +121,8 @@
             "usr/lib/firmware/i915",
             "usr/lib/firmware/intel/vpu",
             "usr/lib/firmware/nvidia",
-            "usr/lib/firmware/xe"
+            "usr/lib/firmware/xe",
+            "usr/lib/firmware/linux-firmware-gpu-version"
         ]
     },
     "migration-manager": {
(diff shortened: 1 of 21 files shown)


stgraber:
@gibmat needs a quick rebase

stgraber:
@gibmat hmm, that's a breaking change that's going to affect anyone who's actively tracking the current version field (like Operations Center).

I'm also not sure how much we want systemd-specific terminology to be exposed vs using something more generic like `image`.

stgraber:
@gibmat

PR #1098, opened 2026-05-14
Add ability to configure serial console(s)

### LINKED ISSUE #940: Configurable text console (serial)
I don't know if this is the right place to report this.

I'm installing IncusOS on a headless system that only has serial console for video output. I couldn't find a way to set the kernel boot parameter `console=ttyS0,115200` to the install image, meaning after BIOS I only get a blank screen.

Is there a way to add that?

It also doesn't seem to have an option to enable it post-install.

gibmat:
Add a `Console` option to the existing system kernel config struct that allows specifying additional consoles for the TUI to use, as well as optionally configuring a baud rate.

A new minimal kernel seed is also defined for cases where setting a console baud rate is required to actually see any output:

```
console:
  - device: /dev/ttyS0
    baud_rate: 11520
```

Closes #940

### DIFF
--- a/doc/reference/seed.md
+++ b/doc/reference/seed.md
@@ -71,6 +71,17 @@ and references Incus' [`InitPreseed` API](https://github.com/lxc/incus/blob/main
 - `preseed`: Additional preseed information to be passed to Incus during
   install.
 
+### `kernel.{json,yml,yaml}`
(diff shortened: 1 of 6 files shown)


gibmat:
Documentation test is failing because of a link to the kernel seed that doesn't exist in the main branch yet.

PR #1135, opened 2026-05-29
Add API endpoint to run signed debug/recovery scripts

### LINKED ISSUE #1033: Add an API to run signed debug/recovery scripts
We currently support running signed debug/recovery scripts from the recovery disk on boot as an emergency mechanism to recover completely broken systems. The signature is checked against the image update signing key.

It would make sense to be able to use that mechanism without the need for a reboot and extra disk being attached the system, so exposing it through a /debug API route should do the trick.

pranav767:
Fixes https://github.com/lxc/incus-os/issues/1033

Creates a debug endpoint which will run signed hotfix script.

- Split the `runHotfix` function to avoid checking and mounting disk in api (Avoid extra disk to be attached)
- Create a new api endpoint in debug which has the hotfix script in body (limits size of script to 1mb) , checks if it is signed, gets CA cert and runs the script.
- Register the new endpoint as `/1.0/debug/:run-script`

### DIFF
--- a/incus-osd/internal/recovery/recovery.go
+++ b/incus-osd/internal/recovery/recovery.go
@@ -102,49 +102,63 @@ func runHotfix(ctx context.Context, updateCA string, mountDir string) error {
 		return nil
 	}
 
-	slog.InfoContext(ctx, "Hotfix script detected, verifying signature")
-
 	f, err := os.Open(filepath.Join(mountDir, "hotfix.sh.sig"))
 	if err != nil {
 		return err
 	}
 	defer f.Close()
 
-	// Validate the signed hotfix script.
-	verified, err := util.VerifySMIME(ctx, updateCA, f)
+	output, err := RunSignedScript(ctx, updateCA, f)
 	if err != nil {
 		return err
+	}
+
+	if output != "" {
+		slog.InfoContext(ctx, "Hotfix script completed", "output", output)
+	}
+
+	return nil
+}
+
+// RunSignedScript verifies and executes a signed hotfix script, returning the script output and any error.
+func RunSignedScript(ctx context.Context, updateCA string, signedScript io.Reader) (string, error) {
+	slog.InfoContext(ctx, "Hotfix script detected, verifying signature")
+
+	// Validate the signed hotfix script.
+	verified, err := util.VerifySMIME(ctx, updateCA, signedScript)
+	if err != nil {
+		return "", err
 	}
 
 	// Write the script contents to a temp file.
 	scriptFile, err := os.CreateTemp("", "")
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	defer os.Remove(scriptFile.Name())
 
 	_, err = scriptFile.WriteString(strings.ReplaceAll(verified.String(), "\r\n", "\n"))
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = scriptFile.Chmod(0o755)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	err = scriptFile.Close()
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	slog.InfoContext(ctx, "Running hotfix script")
 
 	// Run the hotfix script.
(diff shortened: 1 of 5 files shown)


incus-osd/internal/recovery/recovery.go
gibmat: It would be good to move this log line into `RunSignedScript()` as part of the refactoring work.

incus-osd/internal/recovery/recovery.go
```diff
@@ -110,8 +110,14 @@ func runHotfix(ctx context.Context, updateCA string, mountDir string) error {
 	}
 	defer f.Close()
 
+	err = RunSignedScript(ctx, updateCA, f)
+
+	return err
+}
+
+func RunSignedScript(ctx context.Context, updateCA string, signedScript io.Reader) error {
```
gibmat: Exported functions require a comment; not sure why the linter didn't complain about this.

incus-osd/internal/rest/api_debug.go
```diff
@@ -67,7 +69,7 @@ func (*Server) apiDebug(w http.ResponseWriter, r *http.Request) {
 
 	urls := []string{}
 
-	for _, debug := range []string{"log", "processes", "secureboot"} {
+	for _, debug := range []string{"log", "processes", "run-script", "secureboot"} {
```
gibmat: This will also require a small update to one of our end-to-end tests: https://github.com/lxc/incus-os/blob/main/incus-osd/tests/incusos_tests/tests_incusos_api_debug.py#L19..L24

incus-osd/internal/recovery/recovery.go
gibmat: It would be very useful for debugging purposes to capture stdout/stderr and return them regardless of the reported error.

The output could then be logged and additionally returned to the caller if triggered by the end API endpoint. It should be safe to assume it will be unstructured text.

gibmat (changes requested):
Thanks!

It wasn't mentioned in the initial issue, but it would be good to actually capture and return the output when the script is run. This will be especially useful if we ever need to debug when something's gone wrong.

Please split the changes up into multiple commits:

- Refactoring of the `recovery` package code
- Addition of the new REST API
- Update to the debug end-to-end test

Additionally, since this adds a new API endpoint, we need to update the generated docs by running `make update-api`. The changes to `doc/rest-api.yaml` always go in their own commit because of the large, automatically generated changes can make for a messy diff.

incus-osd/internal/rest/api_debug.go
gibmat: Oh, and the endpoint should be `:run-script` to be consistent.

incus-osd/internal/rest/api_debug.go
```diff
@@ -484,3 +486,75 @@ func (*Server) apiDebugSecureBootUpdate(w http.ResponseWriter, r *http.Request)
...
+	if err != nil {
+		_ = response.InternalError(err).Render(w)
+
+		return
+	}
+
+	// Verify the signature and run the script.
+	output, err := recovery.RunSignedScript(r.Context(), updateCA, r.Body)
+	if err != nil {
+		_ = response.InternalError(err).Render(w)
```
gibmat: Especially in the error case we'll want the script output returned. :) I don't know if there's any cleaner way of doing this than something like

```
_ = response.InternalError(errors.New(err.Error() + "\n\n" + output)).Render(w)
```

gibmat (changes requested):
I think your first commit has changes that should belong in the other commits. A little rebasing should get that cleared up; don't be afraid to force-push to the branch.

pranav767:
@gibmat , thanks for the reviews and marking comments resolved, I've updated & fixed changes from your comments.
I hope the commits are good as well.

incus-osd/internal/rest/api_debug.go
pranav767: Updated with above error log showing script output.

gibmat (approved):

PR #1140, opened 2026-05-31
Add other OS version to /1.0 endpoint

### LINKED ISSUE #1136: Report the other installed version of IncusOS
It would be useful for the `/1.0` endpoint to report what the other installed version of IncusOS is, either the prior installed version, or the version of a pending update. It would be easy enough to get this by looking at the UKIs present under `/boot/`.

pranav767:
Resolves https://github.com/lxc/incus-os/issues/1136 , Report other installed version of IncusOS.

`GetUkIVersions()` returns current version, if a second version is present (can be prior version of IncusOS/update that is pending a system reboot) its information will be returned as well.

### DIFF
--- a/doc/rest-api.yaml
+++ b/doc/rest-api.yaml
@@ -10646,6 +10646,7 @@ paths:
                                         machine_id: af94e64e199341b68f10a8eebb828fce
                                         os_name: IncusOS
                                         os_version: "202511041601"
+                                        os_version_alternate: "202511162345"
                                         os_version_next: "202511152230"
                                         system_is_ready: true
                                         uptime: "3237039"
--- a/incus-osd/internal/rest/api_root.go
+++ b/incus-osd/internal/rest/api_root.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/lxc/incus-os/incus-osd/internal/rest/response"
+	"github.com/lxc/incus-os/incus-osd/internal/util"
 )
 
 // If the request contains a X-IncusOS-Proxy header, prepend that to
@@ -110,7 +111,7 @@ func (*Server) apiRoot(w http.ResponseWriter, r *http.Request) {
 //	        metadata:
 //	          type: json
 //	          description: Basic server information
-//	          example: {"environment":{"hostname":"af94e64e-1993-41b6-8f10-a8eebb828fce","machine_id":"af94e64e199341b68f10a8eebb828fce","os_name":"IncusOS","os_version":"202511041601","os_version_next":"202511152230","system_is_ready":true,"uptime":"3237039"}}
+//	          example: {"environment":{"hostname":"af94e64e-1993-41b6-8f10-a8eebb828fce","machine_id":"af94e64e199341b68f10a8eebb828fce","os_name":"IncusOS","os_version":"202511041601","os_version_alternate":"202511162345","os_version_next":"202511152230","system_is_ready":true,"uptime":"3237039"}}
 func (s *Server) apiRoot10(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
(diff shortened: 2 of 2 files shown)


stgraber:
@gibmat does that line up with what you had in mind to pick up the other version? I thought you mentioned scanning through the ESP.

@pranav767 Can you split out the `rest-api.yaml` part of the change to its own commit? We usually have a `doc: Refresh swagger` or similar commit for those kind of automatic changes.

gibmat:
> @gibmat does that line up with what you had in mind to pick up the other version? I thought you mentioned scanning through the ESP.

Yup, this is pretty close to how I would have implemented things. :slightly_smiling_face:

incus-osd/internal/rest/api_root.go
```diff
@@ -151,15 +152,23 @@ func (s *Server) apiRoot10(w http.ResponseWriter, r *http.Request) {
...
-			"os_version":      s.state.OS.RunningRelease,
-			"os_version_next": s.state.OS.NextRelease,
-			"system_is_ready": s.state.OS.SystemIsReady,
-			"uptime":          uptime,
+			"hostname":         s.state.Hostname(),
+			"machine_id":       machineID,
+			"os_name":          s.state.OS.Name,
+			"os_version":       s.state.OS.RunningRelease,
+			"os_version_next":  s.state.OS.NextRelease,
+			"os_version_other": osVersionOther,
```
gibmat: I'd probably call this `alternate_os_version`

incus-osd/internal/rest/api_root.go
```diff
@@ -151,15 +152,23 @@ func (s *Server) apiRoot10(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var osVersionOther string
+
+	ukiVersions, ukiErr := util.GetUKIVersions()
+	if ukiErr == nil {
```
gibmat: We shouldn't ignore an error:

```
	if err != nil {
		_ = response.InternalError(err).Render(w)

		return
	}
```

incus-osd/internal/rest/api_root.go
gibmat: On first boot, there won't be a second UKI present until an update is downloaded. I think we should only report this field if it is not an empty string, so its addition to the returned map should be in a conditional.

gibmat (changes requested):

incus-osd/internal/rest/api_root.go
stgraber: I guess `os_version_alternate` so we can keep the naming pattern.

pranav767: Makes sense, Updated 👍 
The ordering of above map might not be consistent, if we want that can add a struct with `omitempty`

incus-osd/internal/rest/api_root.go
pranav767: Agreed, Updated 👍

pranav767:
> @pranav767 Can you split out the `rest-api.yaml` part of the change to its own commit? We usually have a `doc: Refresh swagger` or similar commit for those kind of automatic changes.

Done :+1:

doc/rest-api.yaml
```diff
@@ -10647,6 +10647,7 @@ paths:
                                         os_name: IncusOS
                                         os_version: "202511041601"
                                         os_version_next: "202511152230"
+                                        os_version_other: "202511162345"
```
stgraber: That shouldn't be in this commit. doc/rest-api.yaml should only be touched in the second commit.

stgraber:
@pranav767 just need to drop doc/rest-api.yaml from the first commit and this should be good to go

pranav767:
> @pranav767 just need to drop doc/rest-api.yaml from the first commit and this should be good to go

My bad, Thought I fixed it with just the second commit, fixed it now :+1:

PR #1147, opened 2026-06-02
incus-osd: Revert to supporting Go 1.25

stgraber:


Makefile
```diff
@@ -233,7 +233,9 @@ test-update-sb-keys:
 .PHONY: update-gomod
 update-gomod:
 	cd incus-osd && go get -t -v -u ./...
-	cd incus-osd && go mod tidy --go=1.26.3
+	cd incus-osd && go get tailscale.com@v1.94.2
+	sed -i "/go-json-experiment/d" incus-osd/go.mod
```
breml, line 237: Wouldn't it be cleaner to do something like this:

```suggestion
	cd incus-osd && go get tailscale.com@v1.94.2 github.com/go-json-experiment/json@ebf49471dced
```

breml (approved):
LGTM, one nit-pick.

Makefile
stgraber: I don't want us to `go get` something that's not one of our dependencies.

This is a weird indirect one that I can't track the actual source for and which `go mod tidy` then does a good job at picking the most recent version that matches its other requirements if we remove the line and run `go mod tidy`.

stgraber: That go-json-experiment/json is pretty odd because according to `go mod graph`, we're directly using it, but we're not and no of the other packages reported as using it where reporting the version that was showing up in go.mod.

It's likely some weird magic because of the upcoming encoding/json/v2 stuff, but we don't use that, nor do we build with it enabled...

Anyway, I found that yanking the line out of go.mod directly, then letting `go mod tidy` run will get us the highest possible version based on our version requirement and keeps it tracked as an indirect dependency, so that seems good enough for now :)

Issue #497, opened 2025-11-07
Offer a full Incus deployment as part of Incus initialization


stgraber:
Currently IncusOS makes it easy to get Incus up and running, including some basic initialization, giving users an initial network and storage pool for their instances.

But to provide the full Incus experience, we'd ideally like to also provide:
 - An OIDC identity provider (Zitadel, Authentik, Keycloak, ...)
 - An OpenFGA instance for authorization
 - A Loki server for logging
 - A Prometheus server for Incus and system metrics
 - A Grafana server for dashboards (integrated with OIDC for login)
 - A reverse proxy (haproxy, traefik, ...) to front everything

This would be all be optionally deployed as part of initialization within an `internal` Incus project and attached to a VXLAN network, making it easy to eventually span this to multiple systems within a cluster (without requiring VLANs or other complex network configuration).

In addition, that same `internal` project could be used to deploy:
 - Linstor controller
 - Ceph services (OSD, MON, MGR, MDS, RGW, ...)
 - OVN services (OVSDB for NorthBound, SouthBound, northd, IC databases, ...)

To make this easy to deploy and manage, we should try to have all of those deployed from OCI images with IncusOS being responsible for generating an initial configuration (environment variables mostly) and then offering the ability to update them by re-building them from the latest version of their respective image.

This would then provide the full Incus experience, all deployed safely on top of IncusOS and with minimal user maintenance required.

(Once we manage to sort all of this one, the next step will be to handle easy clustering, making it as simple as possible to go from a single Incus server with all the aforementioned services to a cluster of 3+ machines with those services getting automatically spread over them)

enriched-tech:
You may want to look at the grafana/otel-lgtm OCI container for these components:

> * A Loki server for logging
> * A Prometheus server for Incus and system metrics
> * A Grafana server for dashboards (integrated with OIDC for login)

That container is easy to setup on Incus, solidly runs for months, and just works.  caveat: I haven't tried integrating Grafana with OIDC .

stgraber:
I think we're more likely to want to run them separately so we have a better path to HA on some of those components. We also don't need the entire set of OTEL components.

Currently I'm looking at:
 - Grafana (dashboard, integrated with Rauthy for authentication and authorization)
 - Prometheus (metrics gathering)
 - Loki (log storage)
 - Alloy (run as a syslog to Loki bridge)

[...]

PR #1159, opened 2026-06-09
Add ability to automatically deploy a Ceph cluster and add it as an Incus storage pool

gibmat:
This adds the capability for IncusOS to automatically deploy a Ceph cluster running from self-hosted OCI application images. Each ODS instance utilizes a LUKS-encrypted raw storage device from the IncusOS server on which it is running.

The logic assumes all Incus cluster members are running on top of IncusOS.

Partially addresses #497

### DIFF
--- a/doc/tutorials/storage-create-ceph-cluster.md
+++ b/doc/tutorials/storage-create-ceph-cluster.md
@@ -1,0 +1,118 @@
+# Creating a Ceph cluster
+
+IncusOS can automate the process of deploying a Ceph cluster. Utilizing OCI container images, IncusOS can self-host a resilient and fault-tolerant Ceph storage cluster.
+
+## Requirements
+
+* An [IncusOS cluster](./incus-cluster.md) consisting of at least three servers:
+
+   * Each server should have a storage pool named `local` available to store Ceph OCI image(s)
+
+   * An Incus-managed common network accessible to all cluster servers with IPv6 connectivity; ideally this network should only be available to cluster servers
+
+   * Optionally, a dedicated project to use when creating the Ceph OCI images and containers
+
+* Each cluster server that will host a Ceph OSD must have an unused raw block device available for the OSD to use. IncusOS will setup LUKS encryption of the raw device before passing it through to the OSD container.
+
+An IncusOS cluster deployed by Operations Center should meet all of these requirements out of the box.
+
+## Initializing the Ceph control plane
+
+The first step is to create the Ceph control plane. The Ceph OCI image will be downloaded and three Ceph containers will be created, each on a different Incus cluster server. The initialization logic also ensures the `incus-ceph` application is installed on each member of the Incus cluster, and that the Ceph service is properly configured.
+
+```
+incus admin os application action incus -d '{"action":"initialize-ceph-cluster","config":{"control_servers":"server01,server02,server03"}}'
+```
+
+### Configuration parameters
+
+* `control_servers`: A comma-separated list of three Incus servers to use when initially deploying the Ceph control plane. If not specified, defaults to the first three Incus cluster servers as reported via the Incus API.
+* `network`: The Incus network that should be utilized by the Ceph cluster. If not specified, defaults to `meshbr0`.
+
+## Adding storage drives (OSDs)
+
+Ceph requires a minimum of three OSDs, and only one OSD can run at a time on a given physical Incus server to help ensure proper replication.
+
+Each OSD is bound to a specific Incus server, so the `--target` parameter must be specified when creating a new OSD.
+
+```
+incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server01
+incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server02
+incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server03
+```
+
+### Configuration parameters
+
+* `device_id`: The device ID of the raw block device to be used by the OSD.
+
+## Updating deployed Ceph version
+
+Because Ceph is deployed using OCI images, it's easy to update or change the running version of Ceph. By default, IncusOS creates the Ceph containers using a major version tag, such as `v20`. As the upstream tag is changed to track the latest release, refreshing and rebuilding the Ceph containers will then automatically pick up that version. It is also possible to specify an explicit tag such as `v21` to bump Ceph to the next major version, `v20.2` to more specifically track a series of Ceph releases, or `v20.1.1` to pin to a specific version.
+
+```{note}
+IncusOS will not automatically remove old, unused Ceph OCI images; this must be done manually.
+```
+
+### Configuration parameters
+
+* `oci_tag`: If specified, update the Ceph containers to use the specific OCI image tag; otherwise, refresh the currently used OCI image and update the Ceph containers if a newer version is available.
+
+```
+incus admin os application action incus-ceph -d '{"action":"refresh-images"}'
+```
+
+## Removing a storage drive (OSD)
+
+If four or more OSDs are available, an OSD can be removed from the Ceph configuration. Because each OSD is bound to a specific Incus server, the `--target` parameter must be specified:
+
+```{warning}
+Removing an OSD via the IncusOS API is equivalent to pulling the power cord of a physical server. Prior to removal of the OSD container, be sure to properly remove the OSD from Ceph's configuration and allow sufficient time for any necessary data migrations to complete.
+```
+
+```
+incus admin os application action incus-ceph -d '{"action":"remove-drive"}' --target server01
+```
+
+## Ceph cluster configuration and status
+
+After the Ceph cluster is initialized, the IncusOS `ceph` service will report basic status information:
+
+```
+$ incus admin os service show ceph
+WARNING: The IncusOS API and configuration is subject to change
+
+config:
+  clusters:
+    ceph:
+      client_config: null
+      fsid: 7fb7b08a-ec49-4ec0-abb0-cd4e9b7089bb
+      keyrings:
+        admin:
+          key: AQD+6U9qzwiAEhAAriOdhjIAN53Oson6D0c1yg==
+      monitors:
+        - fd42:98b5:52c7:673d:1266:6aff:fefa:2cca
+        - fd42:98b5:52c7:673d:1266:6aff:fe4b:2007
+        - fd42:98b5:52c7:673d:1266:6aff:fe0a:310f
+  enabled: true
+state:
+  election_epoch: 22
+  fsid: 7fb7b08a-ec49-4ec0-abb0-cd4e9b7089bb
+  fsmap:
+    btime: 2026-07-09T19:27:42:293081+0000
+    epoch: 5
+  health:
+    status: HEALTH_OK
+
+[snip]
+
+
+```
+
+A minimal amount of configuration state is stored in Incus cluster-wide project configuration using keys with the `user.ceph.*` prefix. This is used to manage the deployment of OSDs at the Incus cluster level:
+
+```
+$ incus project show internal
+config:
+  user.ceph.fsid: 7fb7b08a-ec49-4ec0-abb0-cd4e9b7089bb
+  user.ceph.network: meshbr0
+```
--- a/incus-osd/internal/applications/app_incus_services.go
+++ b/incus-osd/internal/applications/app_incus_services.go
@@ -8,10 +8,25 @@ import (
 	"github.com/lxc/incus/v7/shared/subprocess"
 
 	"github.com/lxc/incus-os/incus-osd/api"
+	"github.com/lxc/incus-os/incus-osd/internal/ceph"
 )
 
 type incusCeph struct {
 	common
+}
+
+// Action runs an application-specific action/task.
+func (*incusCeph) Action(ctx context.Context, data api.ApplicationAction) error {
+	switch data.Action {
+	case "add-drive":
+		return ceph.AddOSD(ctx, data.Config)
+	case "refresh-images":
+		return ceph.RefreshCephOCIImages(ctx, data.Config)
+	case "remove-drive":
+		return ceph.RemoveOSD(ctx)
+	default:
+		return errors.New("unsupported action '" + data.Action + "'")
+	}
 }
 
 func (i *incusCeph) Get(_ context.Context) (any, error) {
--- a/incus-osd/internal/ceph/ceph.go
+++ b/incus-osd/internal/ceph/ceph.go
@@ -1,0 +1,1164 @@
+package ceph
+
+import (
+	"bytes"
+	"context"
+	"embed"
+	"encoding/json"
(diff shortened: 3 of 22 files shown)


stgraber:
As discussed in our call yesterday, a few things to change/add:

 - Deployment should get triggered by modifying the incus application configuration
   - Initially, it's fine to refuse to save a configuration enabling Ceph if the system isn't clustered and the cluster has less than 3 servers
   - Similarly, we should refuse to deploy Ceph if the initial configuration has less than 3 disks with each disk on a different server
   - Adding/removing disks in the configuration will trigger the matching change to the OSD container. If there are no disk, the server doesn't get an OSD container so we'll automatically create or delete the container based on whether disks are being exported
   - The major Ceph version should be in configuration, updating it will trigger a rolling re-deploy of the various
   - All such configuration should be stored in the cluster through user.XYZ configuration keys
 - All interactions with Incus should be done from Go, using the local unix socket and UseTarget() when a specific server needs to be accessed
 - We should implement a general `action` mechanism on applications, such actions for Ceph would include:
   - Apply updates (within the same major version)
 - We shouldn't automatically create an Incus storage pool, this should be left for the user to do, at least until we have more of the full Incus deployment story sorted out

incus-osd/internal/services/service_ceph.go
```diff
@@ -20,12 +23,35 @@ type Ceph struct {
...
-func (n *Ceph) Get(_ context.Context) (any, error) {
+func (n *Ceph) Get(ctx context.Context) (any, error) {
 	// Initialize target list if missing.
 	if n.state.Services.Ceph.Config.Clusters == nil {
 		n.state.Services.Ceph.Config.Clusters = map[string]api.ServiceCephCluster{}
 	}
 
+	// Get current Ceph cluster status. The JSON output is too verbose, while
+	// the plain text output is almost valid yaml (missing quotation of values).
+	output, err := subprocess.RunCommandContext(ctx, "ceph", "status")
```
stgraber: @gibmat I'm not convinced that this correctly handles degraded/broken situations as HEALTH type messages aren't reported as correct YAML typically:
```
ceph stroot@ceph-central01:~# ceph status
  cluster:
    id:     dd7a8436-46ff-4017-9fcb-9ef176409fc5
    health: HEALTH_WARN
            3 daemons have recently crashed
 
  services:
    mon:        3 daemons, quorum ceph-central03,ceph-central01,ceph-central02 (age 7h)
    mgr:        ceph-central01(active, since 7h), standbys: ceph-central02, ceph-central03
    mds:        2/2 daemons up, 1 standby
    osd:        12 osds: 12 up (since 7h), 12 in (since 11d)
    rbd-mirror: 3 daemons active (3 hosts)
 
  data:
    volumes: 1/1 healthy
    pools:   5 pools, 194 pgs
    objects: 1.98M objects, 5.6 TiB
    usage:   16 TiB used, 26 TiB / 42 TiB avail
    pgs:     194 active+clean
 
  io:
    client:   33 MiB/s rd, 1.3 MiB/s wr, 16 op/s rd, 49 op/s wr
```

or:

```
root@ceph-central01:~# ceph status
  cluster:
    id:     3e3ea3d0-cf12-42f3-8648-e6c19e295c6c
    health: HEALTH_ERR
            1 OSD(s) experiencing slow operations in BlueStore
            There are daemons running an older version of ceph
            1 scrub errors
            Possible data damage: 1 pg inconsistent
 
  services:
    mon:        3 daemons, quorum ceph-central01,ceph-central03,ceph-central02 (age 4d)
    mgr:        ceph-central01(active, since 7d), standbys: ceph-central03, ceph-central02
    osd:        16 osds: 16 up (since 2d), 16 in (since 7d)
    rbd-mirror: 3 daemons active (3 hosts)
 
  data:
    pools:   10 pools, 385 pgs
    objects: 1.03M objects, 3.5 TiB
    usage:   10 TiB used, 19 TiB / 29 TiB avail
    pgs:     384 active+clean
             1   active+clean+inconsistent
 
  io:
    client:   3.1 MiB/s rd, 7.3 MiB/s wr, 293 op/s rd, 126 op/s wr
```

or:

```
[root@ceph-central01 ~]# ceph status
  cluster:
    id:     7a96ff17-99e5-4c44-b374-07a63faca559
    health: HEALTH_WARN
            There are daemons running an older version of ceph
 
  services:
    mon:        3 daemons, quorum ceph-central02,ceph-central01,ceph-central03 (age 7d) [leader: ceph-central02]
    mgr:        ceph-central02(active, since 7d), standbys: ceph-central01, ceph-central03
    mds:        1/1 daemons up, 2 standby
    osd:        30 osds: 30 up (since 7d), 30 in (since 7d)
    rbd-mirror: 3 daemons active (3 hosts)
 
  data:
    volumes: 1/1 healthy
    pools:   4 pools, 193 pgs
    objects: 1.07M objects, 3.2 TiB
    usage:   7.4 TiB used, 66 TiB / 74 TiB avail
    pgs:     193 active+clean
 
  io:
    client:   125 KiB/s rd, 4.2 MiB/s wr, 19 op/s rd, 147 op/s wr
```

gibmat: Switched to directly parsing the json output

doc/tutorials/storage-create-ceph-cluster.md
```diff
@@ -0,0 +1,142 @@
...
+
+IncusOS can automate the process of deploying a Ceph cluster. Utilizing OCI container images, IncusOS can self-host a resilient and fault-tolerant Ceph storage cluster.
+
+## Requirements
+
+* An [IncusOS cluster](./incus-cluster.md) consisting of at least three servers:
+
+   * Each server should have a storage pool named `local` available to store Ceph OCI image(s)
+
+   * A common network accessible to all cluster servers with IPv6 connectivity; ideally this network should only be available to cluster servers
```
ykazakov: Should the network automatically assign IPv6 addresses (via DHCPv6 or SLAAC)? Or is it possible to configure static IP addresses? Should the network be managed by Incus or it can be a bridge / VLAN interface created by IncusOS? Specifically, will it be possible to use an isolated network switch for CEPH with just Layer 2 connectivity?

gibmat: > Should the network automatically assign IPv6 addresses (via DHCPv6 or SLAAC)? Or is it possible to configure static IP addresses?

At the moment, yes, the network will need to support automatic IPv6 assignment. Once https://github.com/lxc/incus/pull/3522 lands in Incus, it should be possible to configure specific static IPs if desired.

> Should the network be managed by Incus or it can be a bridge / VLAN interface created by IncusOS? Specifically, will it be possible to use an isolated network switch for CEPH with just Layer 2 connectivity?

It does need to be an Incus-managed network; I'll clarify that in the tutorial.

ykazakov: > Once lxc/incus#3522 lands in Incus, it should be possible to configure specific static IPs if desired.

Nice!

B.t.w., it is possible (now) to assigning a static IP address to an OCI container using `nictype: routed`:
```
devices:
  eth0:
    ipv4.address: 10.30.0.41
    name: eth0
    nictype: routed
    parent: ceph
    type: nic
```

Where `ceph` is a VLAN interface configured in IncusOS:
```
vlans:
  - addresses:
      - 10.30.0.11/24
    id: 30
    name: ceph
    parent: switch
```

ykazakov: > Once lxc/incus#3522 lands in Incus, it should be possible to configure specific static IPs if desired.

Incus 7.2 out, which includes this feature. However, as I understand, it allows setting static IP addresses only for managed bridge networks wit DHCP. This is not particularly useful for CEPH because in a clustered environment running a DHCP servers on each cluster member results in a conflict. For LXC containers, one could set a static IP from within the container without needing a DHCP server or even L3 network.

stgraber: It requires an internal bridged type NIC but it shouldn't require that to be a managed network, so effectively making it work with all IncusOS physical bridges too.

ykazakov: Hmm... I am trying to follow the example on the release notes, but using a physical network:
```
% incus config device set my-nginx eth0 ipv4.address=10.10.10.2/24 ipv4.gateway=10.10.10.1
Error: Invalid devices: Device validation failed for "eth0": Cannot use manually specified ipv4.address when using unmanaged parent bridge
% incus config device show my-nginx     
eth0:
  name: eth0
  network: CEPH
  type: nic
% incus network show CEPH
config:
  vlan: "40"
  volatile.last_state.created: "false"
description: Ceph underlay network
name: CEPH
type: physical 
```

stgraber: Ah, oops, okay, so it's an unrelated check that predates the feature which is kicking in here. Will get that one resolved.

stgraber: Sent a fix for it now

mkosi.images/base/mkosi.conf.d/03-core-packages.conf
```diff
@@ -24,6 +24,7 @@ Packages=
     polkitd
     prometheus-node-exporter
     sanlock
+    skopeo
```
stgraber: Hmm, that shouldn't be needed, you already have it in /opt/incus/bin/

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1268 @@
...
+	//
+	// Save cluster-wide Ceph configuration.
+	//
+
+	server, serverEtag, err := incusClient.GetServer()
+	if err != nil {
+		return err
+	}
+
+	server.Config["user.ceph.project"] = config["project"]
```
stgraber: Hmm, so for now, I think we should try to keep the server config untouched and just assume that we use `internal` for this stuff.

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1268 @@
...
+	// Finalize creation of the new Ceph storage pool.
+	incusClient = incusClient.UseTarget("")
+
+	err = incusClient.CreateStoragePool(incusapi.StoragePoolsPost{
+		Name:   config["pool_name"],
+		Driver: "ceph",
+	})
+	if err != nil {
+		return err
+	}
```
stgraber: I thought we said we wouldn't create a Ceph pool, just put the cluster configuration in place and let the user create what they want.

From what I recall of our conversation, my rationale was that we want to drive this feature through Operations Center and Operations Center is already in charge of enforcement of storage pool configuration through Terraform, so we don't want a conflict here.

Once we land the rest of the IncusOS "full service" deployment, then that logic will take care of the pool creation once Ceph is operational.

incus-osd/internal/applications/app_incus_services.go
```diff
@@ -8,12 +8,29 @@ import (
...
 	common
 }
 
+// Action runs an application-specific action/task.
+func (*incusCeph) Action(ctx context.Context, data api.ApplicationAction) error {
+	switch data.Action {
+	case "add-osd":
+		return ceph.AddOSD(ctx, data.Config)
+	case "create-pool":
+		return ceph.CreateCephPool(ctx, data.Config)
```
stgraber: Okay, so I see it's a separate action, though I'm not sure why we need that at all.

Once Ceph is configured, we should have a functional Ceph cluster and every server should have access to it, so then the user just needs to do the actual:

```
incus storage create my-pool ceph source=my-osd-pool-name
```

And that will have Incus take care of creating the OSD pool and start using it.

incus-osd/internal/applications/app_incus_services.go
```diff
@@ -8,12 +8,29 @@ import (
...
 )
 
 type incusCeph struct {
 	common
 }
 
+// Action runs an application-specific action/task.
+func (*incusCeph) Action(ctx context.Context, data api.ApplicationAction) error {
+	switch data.Action {
+	case "add-osd":
```
stgraber: I'd be tempted to use IncusOS terminology here, so `disk` or `drive` (I don't recall which we use) rather than `osd`.

The term OSD will certainly still be visible in a bunch of places, but I don't want us to be too tied to it given that `ceph-osd` will eventually get phased out with `ceph-crimson` and terminology may change.

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1179 @@
...
+
+		// Ensure the service is enabled.
+		cephService.Config.Enabled = true
+
+		if cephService.Config.Clusters == nil {
+			cephService.Config.Clusters = make(map[string]api.ServiceCephCluster)
+		}
+
+		// Inform the service about our Ceph cluster.
+		cephService.Config.Clusters["ceph"] = api.ServiceCephCluster{
```
stgraber, line 277: So I think we should add a validation step prior to deployment which will go over all the servers in the cluster and check that we don't already have a `ceph` cluster defined in the Ceph service configuration on any of the servers.

We don't want to go stepping on existing configuration.

This will also have the side benefit of confirming that all the servers are online prior to deploying Ceph.

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1179 @@
...
+	//
+
+	project, etag, err := incusClient.GetProject(projectName)
+	if err != nil {
+		return err
+	}
+
+	osds := []osdInfo{} //nolint:prealloc
+
+	if project.Config["user.ceph.osds"] != "" {
```
stgraber: I'm not sure about the `user.ceph.osds` struct, does it hold anything we can't determine directly from Incus and Ceph?

The presence of the OSD instance should be as simple as a name check in GetInstancesNames. What OSD id is on one server can be retrieved from Ceph. What disk is attached to what OSD instance and under what path can be retrieved from Incus.

It feels to me like we could put together that same struct using existing live data rather than having this config key around.

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1179 @@
...
+			"path":      "/etc/ceph/",
+			"dependent": "true",
+		},
+		"var": {
+			"type":      "disk",
+			"pool":      "local",
+			"source":    cephContainerName + "-var",
+			"path":      "/var/lib/ceph/",
+			"dependent": "true",
+		},
```
stgraber, line 525: Instead of using two separate volumes here, I'd make use of two paths within the same volume. Incus should let you create something like `ceph-central01-data` and then use `ceph-central01-data/etc` and `ceph-central01/var` as source paths.

incus-osd/internal/ceph/ceph.go
```diff
@@ -0,0 +1,1179 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	incusClient = incusClient.UseProject(projectName)
+
+	// Grab required configuration files from the initial Ceph server.
+	ret := &clusterConfigFiles{}
+
+	reader, _, err := incusClient.GetInstanceFile(instanceName, "/etc/ceph/ceph.conf")
```
stgraber: GetInstanceFile/UpdateInstanceFile/CreateInstanceFile are the older file API functions. They're fine but not very efficient or flexible.

I'd recommend you replace their use with GetInstanceFileSFTP instead and then use the SFTP connection to go through all the file operations.

stgraber:
Sorry for taking so long to review again. I was definitely using the fact that you were off at Debconf to go focus on other things first ;)

This looks pretty good and very close to ready. Just a few notes above which hopefully make sense and can be done pretty quickly.

incus-osd/internal/ceph/ceph.go
gibmat: Will want to make sure https://github.com/lxc/incus/pull/3740 is included in the Incus feature/LTS zabbly package builds.

PR #1176, opened 2026-06-23
Fixes

gibmat:
* incus-osd/tests: Ensure /usr/sbin is in path for running commands
* cli: Add an `application edit` command


gibmat:
Documentation check failing due to rate-limiting of the link checker.

PR #1178, opened 2026-06-23
Align LLM policy with Incus

stgraber:


gibmat (approved):

gibmat:
A couple of new words need to be added to `doc/.wordlist.txt` :)

PR #1179, opened 2026-06-23
Add support for configuring zram-backed swap

### LINKED ISSUE #1157: Option to use zram swap
In view of the ongoing DRAM shortage, it would be very useful to be able to configure a zram swap device, as the swap is quite small on IncusOS and there does not seem to be a way to increase it. By allocating a large zram device for swap we can effectively increase the available memory at the cost of some CPU cycles, which could be a worthwhile tradeoff for many use cases where the server runs a lot of services that are largely idle or IO bound.

Reference: https://wiki.archlinux.org/title/Zram

It is the default setup on many distros, like [Fedora](https://www.fedoraproject.org/wiki/Changes/SwapOnZRAM) and Arch when using the installer. At a minimum it is much faster than any type of disk swap currently being used.

Suggested configuration options:
- zram_compression_algorithm
- zram_size

P.S. It may also be useful to use this facility as a generic block device by creating new zram devices for use in VMs.

gibmat:
Add a new kernel memory configuration option `zram_swap_size` that enables configuring zram-backed swap of the specified size. A value > 0 will (re)create the zram swap device, while a value of 0 will remove it. The configuration is persisted in the IncusOS state, and will automatically apply when `incus-osd` starts up.

The compression algorithm is fixed to zstd, which is the only one supported out-of-the-box in the IncusOS kernel. I think that's fine for now; if there's really a pressing need we can look at enabling additional kernel modules at a later point.

Closes #1157

### DIFF
--- a/incus-osd/internal/kernel/kernel.go
+++ b/incus-osd/internal/kernel/kernel.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/lxc/incus/v7/shared/subprocess"
+	"github.com/lxc/incus/v7/shared/units"
 
 	"github.com/lxc/incus-os/incus-osd/api"
 	"github.com/lxc/incus-os/incus-osd/internal/systemd"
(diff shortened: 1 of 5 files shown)


incus-osd/internal/kernel/kernel.go
```diff
@@ -38,9 +42,114 @@ func ApplyKernelConfiguration(ctx context.Context, config api.SystemKernelConfig
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// The zramctl command is a bit too flaky and frequently returns "Device or resource busy"
+// errors. To handle this, try each command up to ten times with brief sleeps in between
+// before fully failing.
```
rugk, line 137: Hmm, does IncusOS use systemd? If so, why not use `systemd zram-generator`? IMHO [Fedora has written up why they choose this quite well](https://fedoraproject.org/wiki/Changes/SwapOnZRAM#Why_systemd_zram-generator?):
> It's the most upstream implementation to date, is fast and lightweight. The zram-generator uses existing systemd infrastructure to setup the zram block device, format it as swap, and swapon - all during early boot.
> Converging on one implementation avoids user confusion. And while the alternatives are nice and work fine, a systemd generator is particularly well suited for this use case compared to a systemd service unit.
> Also, it's an reference implementation of a system generator written in Rust.

rugk (commented):
Thanks, looks great!

Also, BTW, for anyone wondering what size to use [Fedora increased zram to the maximum of RAM with or 8GB, whichever is smaller. (`zram-fraction=1.0, max-zram-size=8192`),

Also, `zramctl` seems to show statistics about the used compression ratio etc. Maybe that can be exposed in IncusOS, too (in some `state:` YAML?)?

incus-osd/internal/kernel/kernel.go
stgraber: We want it to be opt in and to have a configurable size. Doing that with the generator would require the creation of an early boot override and then generation of the config in /etc by incus-osd followed by re-running the generator (daemon-reload) and restarting the generated systemd units. 

That all ends up being more logic and complexity than just driving the thing ourselves.

gibmat:
Added some basic zram statistics as part of https://github.com/lxc/incus-os/pull/1181.

PR #1183, opened 2026-06-25
mkosi.packages/initrd-utils: Run fsck on the root partition immediately after cryptsetup completes

### LINKED ISSUE #1101: Enable `systemd-fsck-root.service`
We should make sure `systemd-fsck-root.service` is properly activated and scanning the root (ext4) partition on each boot.

gibmat:
Ideally we would let systemd-fsck-root.service to do this for us, but because we rely on swtpm storing state in /boot/, dependency ordering prevents that service from running after the root partition is unlocked but before it is mounted. Instead, immediately run fsck ourselves from the cryptsetup service.

Closes #1101

### DIFF
--- a/mkosi.packages/initrd-utils/initrd-fsck-root.sh
+++ b/mkosi.packages/initrd-utils/initrd-fsck-root.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# Ideally systemd-fsck-root.service would check the root partition automatically, but because of
+# how we are relying on swtpm storing state in /boot/, the unencrypted root device won't be available.
+# Attempting to add additional service ordering dependencies results in a dependency cycle, so we mimic
+# what that service does immediately following the successful unlocking of the root device.
+
+/usr/sbin/fsck.ext4 -p /dev/mapper/root
+ret=$?
+
+# Perform a couple actions based on the exit code of fsck. If it failed to repair the file system, don't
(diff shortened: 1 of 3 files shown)


mkosi.packages/initrd-utils/initrd-fsck-root.sh
```diff
@@ -0,0 +1,23 @@
...
+    mkdir -p /run/initramfs/
+    touch /run/initramfs/fsck-root
+
+    exit
+elif [ $ret = 2 ]; then
+    # A system reboot is required
+    systemctl reboot
+else
+    # Propagate the error
+    exit $ret
```
stgraber: @gibmat what the consequence there?

If we fail to boot because of this, that'd be a bit problematic as there's not much the user can really do. Should we not instead record the failure and output and attempt to boot anyway so incus-osd can report it?

gibmat: We could metaphorically kick the can down the road by not exiting with an error. If fsck isn't able to repair the issues, I'd guess the filesystem would refuse to mount, which would then likely cause `sysroot.mount` to fail shortly later in the initrd startup.

stgraber: Yeah, that may be best, just to avoid a weird initrd-only failure breaking a bunch of systems.

stgraber (approved):

PR #1191, opened 2026-06-30
Add a `security` seed

gibmat:
Add a new `security` seed to allow configuration of custom CAs during first boot.

Also updated expected Java paths for the new version of Linstor which bumped to Java 25.

### DIFF
--- a/mkosi.images/base/mkosi.extra/usr/lib/systemd/system/incus-osd.service
+++ b/mkosi.images/base/mkosi.extra/usr/lib/systemd/system/incus-osd.service
@@ -13,7 +13,7 @@ Environment=TERM=xterm-256color
 KillMode=process
 TimeoutStartSec=30s
 TimeoutStopSec=30s
-Restart=on-failure
+Restart=always
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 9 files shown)


mkosi.images/base/mkosi.extra/usr/lib/systemd/system/incus-osd.service
```diff
@@ -13,7 +13,7 @@ Environment=TERM=xterm-256color
 KillMode=process
 TimeoutStartSec=30s
 TimeoutStopSec=30s
-Restart=on-failure
+Restart=always
```
stgraber, line 16: What about when it exits cleanly as it does on poweroff and reboot?

gibmat: In those cases, the system will already be shutting down, since we've called `systemctl poweroff` or `systemctl reboot` before exiting. And there's nowhere else in the code that we call `os.Exit(0)`.

stgraber: Ah, fair, systemd won't be respawning the service if the system is shutting down.

PR #1197, opened 2026-07-05
incus-osd: Include subsystems info when getting targets

evertonen:
I used the IncusOS image downloader WEB UI wizard to create an installation image, but the installer was not being able to find and pick the specific nvme when the "Target drive bus" field was filled in with "nvme".

I believe this patch should fix the issue.

Example:
```
$ lsblk -N -iJnpb -e 1,2 -o KNAME,ID_LINK,SIZE 
{
   "blockdevices": [
      {
         "kname": "/dev/nvme0n1",
         "id-link": "nvme-eui.001b448b491b9b6c",
         "size": 512110190592
      }
   ]
}
$ 
$ lsblk -N -iJnpb -e 1,2 -o KNAME,ID_LINK,SIZE,SUBSYSTEMS
{
   "blockdevices": [
      {
         "kname": "/dev/nvme0n1",
         "id-link": "nvme-eui.001b448b491b9b6c",
         "size": 512110190592,
         "subsystems": "block:nvme:pci"
      }
   ]
}
$ 
```

### DIFF
--- a/incus-osd/internal/install/install.go
+++ b/incus-osd/internal/install/install.go
@@ -456,7 +456,7 @@ func getAllTargets(ctx context.Context, sourceDevice string) ([]storage.BlockDev
 	// Get NVME drives first.
 	nvmeTargets := storage.LsblkOutput{}
 
-	output, err := subprocess.RunCommandContext(ctx, "lsblk", "-N", "-iJnpb", "-e", "1,2", "-o", "KNAME,ID_LINK,SIZE")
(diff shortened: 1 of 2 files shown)


stgraber:
Hmm, pretty sure I used the `nvme` bus pretty recently. Certainly possible that we regressed something though.

@gibmat

evertonen:
I've updated the example using my laptop. But the device where I flashed Incus OS has 2 NVMEs and I ended up using just the "Use the smallest matching driver as boot drive" option with all other fields blank.

gibmat (approved):
Yeah... oops this wasn't actually working, although the unit tests are! 😆 We had extended the `lsblk` arguments in the storage API, but neglected to do so in the install logic.

I've added a small test to verify the bus selector works in the install seed, as well as catch any regressions in the future.

PR #1202, opened 2026-07-07
Improve S/MIME validation checks

gibmat:
Currently, IncusOS verifies S/MIME content by using the built-in root CA as a trust anchor and relies on the S/MIME message to include a proper certificate chain consisting of an intermediate CA and the actual signing certificate. This works fine, but would allow for a compromised intermediate CA to issue certificates and make signatures for sources of data that it shouldn't. This PR introduces additional checks so that only an expected CA can be used to sign certain updates:

* The `images` provider will rely json indices being signed by the Update CA (current behavior)
* Any hotfix scripts must be signed by the Support CA (this will be the Update CA for IncusOS, but different for HypervisorOS)
* Any recovery update json indices must be signed by the Update CA (current behavior)

### DIFF
--- a/.github/workflows/certificate-e2e.yml
+++ b/.github/workflows/certificate-e2e.yml
@@ -1,0 +1,93 @@
+name: Tests
+on:
+  schedule:
+  - cron: "0 0 * * 0"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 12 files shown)


stgraber:
@gibmat I would only allow the support CA for hotfixes, whether from a rescue disk or from API. Updates should always exclusively be checked against the update CA.

PR #1205, opened 2026-07-08
Misc fixes

### LINKED ISSUE #1206: Duplicate MAC validation failing for bonds
Given the following simplified network config (enough to fail validation):
```
bonds:
  - name: bond1
    members:
      - 10:10:10:10:10:01
      - 10:10:10:10:10:02
    
  - name: bond2
    members:
      - 10:10:10:10:10:03
      - 10:10:10:10:10:04
```
When that gets parsed, `bond.hwaddr` is an empty string

When its validated here the empty string gets added to the slice.
- https://github.com/gibmat/incus-os/blob/829415edc9ad3ff4242e209567ceef8f1f29ad87/incus-osd/internal/systemd/networkd.go#L228

And on the next loop the second bond has `bond.hwaddr` as an empty string and results in a duplicate MAC error.

For bonds, I guess we need to validate based on a combination of the `hwaddr` if its set and also the `members`?

gibmat:
* Clear temporary swtpm state before initialization during install
* Ignore iDRAC virtual devices 
* Fix bond duplicate MAC check and also check member MACs

Closes #1206

### DIFF
--- a/incus-osd/internal/systemd/networkd.go
+++ b/incus-osd/internal/systemd/networkd.go
@@ -225,12 +225,36 @@ func ValidateNetworkConfiguration(networkCfg *api.SystemNetworkConfig, requireVa
 			return errors.New("duplicate interface/bond/vlan/wireguard name: " + bond.Name)
 		}
 
-		if slices.Contains(macs, bond.Hwaddr) {
-			return errors.New("duplicate MAC address: " + bond.Hwaddr)
+		names = append(names, bond.Name)
+
+		// A bond configuration may not explicitly define a MAC, so only check if one is present.
+		if bond.Hwaddr != "" {
+			if slices.Contains(macs, bond.Hwaddr) {
+				return errors.New("duplicate MAC address: " + bond.Hwaddr)
+			}
+
+			macs = append(macs, bond.Hwaddr)
 		}
 
-		names = append(names, bond.Name)
-		macs = append(macs, bond.Hwaddr)
+		// Check that each bond member has a unique MAC, with the exception that the
+		// bond may be configured with the same MAC as one of its members.
+		numMembersWithBondMac := 0
+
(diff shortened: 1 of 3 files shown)


incus-osd/internal/systemd/networkd.go
```diff
@@ -225,12 +225,15 @@ func ValidateNetworkConfiguration(networkCfg *api.SystemNetworkConfig, requireVa
...
 		}
 
-		if slices.Contains(macs, bond.Hwaddr) {
-			return errors.New("duplicate MAC address: " + bond.Hwaddr)
-		}
-
 		names = append(names, bond.Name)
-		macs = append(macs, bond.Hwaddr)
+
+		if bond.Hwaddr != "" {
```
stgraber, line 231: @gibmat we also need to check for duplicated MACs within the bond itself

incus-osd/internal/systemd/networkd.go
```diff
@@ -225,12 +226,27 @@ func ValidateNetworkConfiguration(networkCfg *api.SystemNetworkConfig, requireVa
...
+			macs = append(macs, bond.Hwaddr)
+			bondMacs = append(bondMacs, bond.Hwaddr)
 		}
 
-		names = append(names, bond.Name)
-		macs = append(macs, bond.Hwaddr)
+		// Check that each bond member has a unique MAC, with the exception that the
+		// bond may be configured with the same MAC as one of its members.
+		for _, memberMAC := range bond.Members {
+			if slices.Contains(macs, memberMAC) && !slices.Contains(bondMacs, memberMAC) {
```
stgraber: I don't think we want to allow one bond to use the same hwaddr as a member of another bond, which this now allows.

So instead of tracking `bondMacs` I think we should instead allow for `memberMAC == bond.Hwaddr` instead.

PR #1222, opened 2026-07-22
Push MachineID and SystemUUID to Operations Center

stgraber:


stgraber:
@breml

incus-osd/internal/state/struct.go
```diff
@@ -110,6 +110,20 @@ func (*State) MachineID() (string, error) {
...
+	productUUID, err := os.ReadFile("/sys/class/dmi/id/product_uuid")
+	if err != nil {
+		return "", err
+	}
+
+	if len(productUUID) != 37 {
+		return "", errors.New("invalid length for a UUID")
+	}
+
+	return strings.TrimSpace(string(productUUID)), nil
```
breml, line 124: I would trim the productUUID and only then check for the length, since if it passes the length check, then there is nothing to trim anyway.

breml:
@stgraber LGTM, but I left one comment about the length check, which you might want to consider.

incus-osd/internal/state/struct.go
stgraber: I was going for minimal amount of change, basically lifting the logic as it was and aligning it with the machine-id logic.

TrimSpace here is actually abused as a way to remove the expected trailing line break rather than remove any other unexpected spaces.

PR #1240, opened 2026-07-30
gpu-support docs: clarify this excludes out of tree drivers

woodsb02:
Update GPU support documentation to reflect correct driver names, and add a note to clarify out of tree Nvidia drivers are not included.

### DIFF
--- a/doc/reference/applications/gpu-support.md
+++ b/doc/reference/applications/gpu-support.md
@@ -2,6 +2,10 @@
 
 The GPU support (`gpu-support`) application includes the required firmware files for GPUs from one of:
 
-* `amd`
-* `nvidia`
-* `intel`
+* `amdgpu`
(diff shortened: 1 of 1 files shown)


stgraber:
That change is incorrect. `gpu-support` does NOT include ANY driver at all.
It ONLY includes firmware files as its documentation currently describes.

Those firmware files are not driver specific, so it's completely incorrect to list them as being for specific drivers. They are the exact same firmware files you need if you were running the proprietary nvidia or amdgpu-pro drivers.

Kernel modules are always part of the base image, never leaf images like gpu-support.

woodsb02:
Ok - i think this is really good information that should be captured in the docs, so that people don’t get confused and are only able to find answers on the forum. I’ve tried again with #1241

PR #1241, opened 2026-07-30
gpu-support docs: clarify this does not include drivers

woodsb02:
Clarify gpu-support includes only firmware files and not drivers. Added note about kernel modules and Nvidia drivers.

### DIFF
--- a/doc/reference/applications/gpu-support.md
+++ b/doc/reference/applications/gpu-support.md
@@ -5,3 +5,8 @@ The GPU support (`gpu-support`) application includes the required firmware files
 * `amd`
 * `nvidia`
 * `intel`
+
+```{note}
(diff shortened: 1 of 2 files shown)


doc/reference/applications/gpu-support.md
```diff
@@ -5,3 +5,9 @@ The GPU support (`gpu-support`) application includes the required firmware files
 * `amd`
 * `nvidia`
 * `intel`
+
+```{note}
+gpu-support only includes firmware files - it does not include any drivers.
```
stgraber: Nvidia needs to be added to the word list and gpu-support needs to always be escaped as `gpu-support`

stgraber:
Missing the Signed-off-by line in the commit message.

stgraber (approved):

PR #1250, opened 2026-08-03
Add openfga-sync

stgraber:


gibmat (approved):
I'd move the go.mod commit earlier in the commits before any code actually imports the openfga-sync config, just in case someday we're doing a `git bisect` and happen to land somewhere in this PR commit series.

stgraber:
> I'd move the go.mod commit earlier in the commits before any code actually imports the openfga-sync config, just in case someday we're doing a `git bisect` and happen to land somewhere in this PR commit series.

done!

PR #1257, opened 2026-08-05
incus-osd/app/openfga: Only override sync target URL/token

stgraber:


stgraber:
We were accidentally masking other user-configurable flags like Authoritative.

gibmat (approved):

Issue #1258, opened 2026-08-05
IncusOS won't import externally-created, unencrypted zpools


hosaka:
This happened twice today with an `202608021451` update. Once on my Proxomox server at work and once on my homelab setup.

I have 3 nodes, each with a spare NVME for use with linstor, all nodes were configured and working with linstor, node-1 is hosting a linstor-controller in a debian-13 container. It looks like the `nvme_pool` simply disappeared after the update. Before updating to `202608021451`, from the linstor container:

```shell
# linstor storage-pool list
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ StoragePool          │ Node   │ Driver   │ PoolName  │ FreeCapacity │ TotalCapacity │ CanSnapshots │ State │ SharedName                  │
╞══════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╡
│ DfltDisklessStorPool │ node-1 │ DISKLESS │           │              │               │ False        │ Ok    │ node-1;DfltDisklessStorPool │
│ DfltDisklessStorPool │ node-2 │ DISKLESS │           │              │               │ False        │ Ok    │ node-2;DfltDisklessStorPool │
│ DfltDisklessStorPool │ node-3 │ DISKLESS │           │              │               │ False        │ Ok    │ node-3;DfltDisklessStorPool │
│ nvme_pool            │ node-1 │ ZFS_THIN │ nvme_pool │   230.62 GiB │       238 GiB │ True         │ Ok    │ node-1;nvme_pool            │
│ nvme_pool            │ node-2 │ ZFS_THIN │ nvme_pool │   230.62 GiB │       238 GiB │ True         │ Ok    │ node-2;nvme_pool            │
│ nvme_pool            │ node-3 │ ZFS_THIN │ nvme_pool │   230.62 GiB │       238 GiB │ True         │ Ok    │ node-3;nvme_pool            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

After applying the update, immediately node-1 showed up with an Error state:
```shell
# linstor storage-pool list
╭──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ StoragePool          │ Node   │ Driver   │ PoolName  │ FreeCapacity │ TotalCapacity │ CanSnapshots │ State │ SharedName                  │
[...]

PR #1259, opened 2026-08-05
Allow importing of unencrypted storage pools

gibmat:
Extend the storage API to report if a storage pool is managed by IncusOS or not, and make the lack of an encryption key more obvious.

This also changes logic to allow the importing of unencrypted storage pools. This is **strongly** discouraged, and each unencrypted storage pool must be manually imported via the API after IncusOS starts up.

Partly addresses #1258

### DIFF
--- a/incus-osd/api/system_storage.go
+++ b/incus-osd/api/system_storage.go
@@ -37,6 +37,7 @@ type SystemStoragePool struct {
 	Special *SystemStoragePoolSpecial `json:"special,omitempty" yaml:"special"`
 
 	// Read-only fields returned from the server with additional pool information.
+	Managed                   bool                          `json:"managed"                       yaml:"managed"`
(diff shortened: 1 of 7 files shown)


incus-osd/api/system_storage.go
```diff
@@ -37,6 +37,7 @@ type SystemStoragePool struct {
 	Special *SystemStoragePoolSpecial `json:"special,omitempty" yaml:"special"`
 
 	// Read-only fields returned from the server with additional pool information.
+	ManagedByIncusOS          bool                          `json:"managed_by_incusos"            yaml:"managed_by_incusos"`
```
stgraber: I'd just name the flag `Managed` so we don't end up putting the specific OS name in the variable (which would then get confusing with HypervisorOS).

stgraber (approved):

PR #1264, opened 2026-08-06
incus-osd/image-customizer: Reject network seeds with a "config" key

stgraber:


gibmat (commented):
I feel like it would be better and more robust to actually attempt to decode each seed (network or otherwise) in conjunction with setting `DisallowUnknownFields()`. That should catch any seeds that would cause issues later when actually configuring the system.

stgraber:
@gibmat switched to a strict JSON parser. The unpacking target was already correct.

gibmat (approved):

PR #1284, opened 2026-08-14
incus-osd/rest: Add internal endpoint exposing the Secure Boot certif…

### LINKED ISSUE #1220: Add internal REST API endpoint that exposes the internal SecureBoot keys struct
That's for Operations Center to consume so it can then provision the Redfish BMCs automatically.
We want to expose this struct as JSON: https://github.com/lxc/incus-os/blob/main/incus-osd/certs/certs.go#L33C1-L38C2

stgraber:
…icates

Closes #1220

### DIFF
--- a/incus-osd/api/internal_secure_boot.go
+++ b/incus-osd/api/internal_secure_boot.go
@@ -1,0 +1,9 @@
+package api
+
+// InternalSecureBootCertificates holds the PEM encoded Secure Boot certificates used by the system.
+type InternalSecureBootCertificates struct {
+	PK  string   `json:"pk"  yaml:"pk"`
+	KEK []string `json:"kek" yaml:"kek"`
+	DB  []string `json:"db"  yaml:"db"`
+	DBX []string `json:"dbx" yaml:"dbx"`
+}
--- a/incus-osd/certs/certs.go
+++ b/incus-osd/certs/certs.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"os"
 	"strings"
+
+	"github.com/lxc/incus-os/incus-osd/api"
 )
 
 //go:embed files/*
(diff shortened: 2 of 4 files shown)


incus-osd/api/secure_boot.go
```diff
@@ -0,0 +1,9 @@
+package api
+
+// SecureBootCertificates holds the PEM encoded Secure Boot certificates used by the system.
+type SecureBootCertificates struct {
```
gibmat: I'd really like to see if we can find a way to not end up essentially duplicating the same struct that's currently in the `certs` package: https://github.com/lxc/incus-os/blob/main/incus-osd/certs/certs.go#L33. Maybe this is the easiest way to make the SecureBoot certs available via the API, but it also introduces the need to keep two different structs in sync. Admittedly, they're unlikely to change, but that also means we're pretty like to miss updating one and not the other. 🙂

gibmat (changes requested):

incus-osd/api/secure_boot.go
stgraber: Hmm, yeah, I'll look into options there.

A few options I can think of:

1. We have the public struct use x509 directly with marshaller and unmarshallers for JSON and YAML (may be a bit fragile)
2. We have the internal struct has a ToAPI() function which returns the public struct (easy, lines up with Incus DB logic)
3. We move everything to the public struct and add functions to it to get things as x509

stgraber:
@gibmat went with the ToAPI() approach to this. Also renamed things a bit to make it clear it's for the internal API.

I think it's still useful to have those structs available so @breml doesn't need to duplicate them in Operations Center, but we also don't want them to look like a guaranteed stable API.

gibmat (approved):

PR #1288, opened 2026-08-14
Add services seed

stgraber:


incus-osd/cmd/incus-osd/main.go
```diff
@@ -738,6 +738,25 @@ func startup(ctx context.Context, s *state.State) error { //nolint:revive
...
+		servicesSeed, err := seed.GetServices(ctx)
+		if err != nil && !seed.IsMissing(err) {
+			return errors.New("unable to parse services seed: " + err.Error())
+		}
+
+		if servicesSeed != nil {
+			services.ApplySeed(ctx, s, servicesSeed)
+		}
+
+		s.ServicesSeedApplied = true
```
gibmat, line 752: I suppose we can't really do what we do with applications applying their seed configuration in `Initialize()`... This isn't too bad, but does add another field to state.

doc/reference/seed.md
```diff
@@ -152,6 +152,30 @@ The "debug" provider is not intended for general use, and should only be used to
...
+version: "1"
+tailscale:
+  enabled: true
+  auth_key: tskey-auth-XXXXXXXXXXXX
+lvm:
+  enabled: true
+```
+
+The seed configuration is applied on first boot, before any of the services are
+started. Services requiring an application that isn't installed yet (`ceph` and
```
gibmat, line 170: So let's remove `Ceph` and `Linstor` from the seed struct, so if someone skips reading the documentation they're not confused about why their service configuration isn't applied. 🙂

gibmat (changes requested):

doc/reference/seed.md
stgraber: Good point, done!

gibmat (approved):

PR #1293, opened 2026-08-16
Implement kernel debug API

stgraber:


incus-osd/internal/kernel/debug.go
```diff
@@ -0,0 +1,148 @@
...
+	info.Modules, err = getModules()
+	if err != nil {
+		return nil, err
+	}
+
+	return info, nil
+}
+
+// cpuBaseline determines the x86_64 micro-architecture baseline from the CPU flags.
+func cpuBaseline(cpuinfo string) string {
```
gibmat: I feel like it would be easier to just check the output of running `/lib64/ld-linux-x86-64.so.2 --help`, rather than trying to parse specific architecture flags.

incus-osd/internal/rest/api_debug.go
```diff
@@ -68,14 +69,65 @@ func (*Server) apiDebug(w http.ResponseWriter, r *http.Request) {
 
 	urls := []string{}
 
-	for _, debug := range []string{"log", "processes", ":run-script", "secureboot"} {
+	for _, debug := range []string{"kernel", "log", "processes", ":run-script", "secureboot"} {
```
gibmat, line 72: Will also have to update the test in `tests_incusos_api_debug.py`

gibmat (changes requested):

incus-osd/internal/kernel/debug.go
stgraber: Switched over to that.

incus-osd/internal/rest/api_debug.go
stgraber: Updated

stgraber:
@gibmat I think that's those comments addressed

gibmat (approved):

incus-osd/internal/kernel/debug.go
```diff
@@ -0,0 +1,113 @@
...
+
+		name := fields[0]
+
+		_, ok := dependencies[name]
+		if !ok {
+			dependencies[name] = []string{}
+		}
+
+		// The third field holds the module's reference count.
+		inUse[name] = fields[2] != "0"
```
mihalicyn, line 87: I'm afraid that this is not what we want, because for example for `veth` module on my system I'm getting `0`, while `veth` is obviously used by a running LXC container. Many Linux kernel modules are not calling `module_get` explicitly so this refcounter doesn't really make a much of sense. If you do `rmmod veth` it will go and remove veth devices everywhere in the system... i.e. there is no refcounting or tracking issue and kernel knows what to do because `veth_exit() -> rtnl_link_unregister() -> __rtnl_kill_links()` will be triggered upon module unload.

mihalicyn: Another thing is that it is better to rely on `ls -l /sys/module | awk '{ print $9; }'` rather than `/proc/modules`, because `/proc/modules` only shows only what was compiled and loaded as a module, while `/sys/module` also shows what could be a module but was chosen to be a built-in. We can have a separate category for this, or just list all this as "modules", IMHO.

stgraber: But for built-in, it will all get listed whether the system needs them or not, so it's not as useful to us since we can make that list just by parsing the kconfig for any =y which also supports =m

Issue #1260, opened 2026-08-05
Give better insight into root disk usage


gibmat:
Right now there's really no way to get the current root disk usage via the IncusOS API. We should add at a minimum a way to get that info.

Also, it might be worth adding one or more actions to help free disk space: clear all logs older than X days, selectively remove old versions of sysext images, etc. Since there's no remote shell access, the user can't take any action on their own to free up disk space.

PR #1295, opened 2026-08-18
Root partition usage, CPU governor, SR-IOV devices and network sysctls

stgraber:


stgraber:
Bit of a mix, though most of them touch the same area (kernel config).

incus-osd/api/system_storage.go
```diff
@@ -10,8 +10,16 @@ type SystemStorageConfig struct {
...
-	Pools  []SystemStoragePool  `json:"pools"  yaml:"pools"`
+	Drives        []SystemStorageDrive       `json:"drives"         yaml:"drives"`
+	Pools         []SystemStoragePool        `json:"pools"          yaml:"pools"`
+	RootPartition SystemStorageRootPartition `json:"root_partition" yaml:"root_partition"`
+}
+
+// SystemStorageRootPartition defines a struct that holds usage information about the root ("/") partition.
+type SystemStorageRootPartition struct {
+	SizeInBytes      int `json:"size_in_bytes"      yaml:"size_in_bytes"`
+	UsedInBytes      int `json:"used_in_bytes"      yaml:"used_in_bytes"`
```
gibmat: It seems redundant to report both used and available space, since unlike ZFS with ext4 it's trivial to figure out one from the other if you know the partition's size.

incus-osd/api/system_storage.go
```diff
@@ -10,8 +10,16 @@ type SystemStorageConfig struct {
...
 type SystemStorageState struct {
-	Drives []SystemStorageDrive `json:"drives" yaml:"drives"`
-	Pools  []SystemStoragePool  `json:"pools"  yaml:"pools"`
+	Drives        []SystemStorageDrive       `json:"drives"         yaml:"drives"`
+	Pools         []SystemStoragePool        `json:"pools"          yaml:"pools"`
+	RootPartition SystemStorageRootPartition `json:"root_partition" yaml:"root_partition"`
+}
+
+// SystemStorageRootPartition defines a struct that holds usage information about the root ("/") partition.
+type SystemStorageRootPartition struct {
```
gibmat, line 19: Do we want to close #1260 (ignoring the second half of the issue), or move the cleanup tasks to a new issue?

incus-osd/cmd/incus-osd/main.go
```diff
@@ -565,6 +565,14 @@ func startup(ctx context.Context, s *state.State) error { //nolint:revive
 		}
 	}
 
+	// Create SR-IOV virtual functions, if configured.
```
gibmat: We're getting enough start-up kernel configuration updates that maybe it's better to just call `kernel.ApplyKernelConfiguration()`?

gibmat (changes requested):

incus-osd/api/system_storage.go
stgraber: true, I guess we can stick to available given that's the most useful to a user?

incus-osd/api/system_storage.go
stgraber: Hmm, I can add a :cleanup-root POST route which triggers our current cleanup logic.

incus-osd/api/system_storage.go
gibmat: Yeah, that makes sense.

incus-osd/api/system_storage.go
gibmat: But I think we'll want to expand what sort of cleanup can be performed on the root partition, since there's no shell access to facilitate reclaiming space. Probably the most notable would be removing older versions of sysext images that aren't in use.

stgraber:
@gibmat that should address the review comments so far :)

gibmat (approved):

PR #1321, opened 2026-08-27
doc: Clarify recovery key entry

stgraber:


stgraber:
@breml

doc/reference/recovery.md
```diff
@@ -44,6 +44,12 @@ will need to manually provide an encryption recovery passphrase. After IncusOS s
 up, you can use the API to force-reset the TPM encryption bindings which should allow
 automatic decryption of the install drive at boot time.
 
+```{note}
+When manually entering the recovery, keep in mind that the keyboard
+layout is a traditional United States layout and that the recovery key
```
breml, line 49: Maybe also mention, that the BMC might emulate a different keyboard layout (as we have observed our selves on the DELL iDRAC). With the prompt not showing the actual letter being entered, it is just guessing, what actually end up in there.

gibmat (approved):

PR #1327, opened 2026-08-28
Implement ZFS trim

stgraber:


incus-osd/internal/zfs/zfs.go
```diff
@@ -1087,6 +1089,73 @@ func ScrubAllPools(ctx context.Context) error {
...
+// TrimAllPools trims all pools in the system sequentially, blocking until the trim is complete.
+func TrimAllPools(ctx context.Context) error {
+	_, pools, err := storage.GetStorageInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Trim every pool sequentially.
+	for _, pool := range pools {
+		slog.InfoContext(ctx, "Trimming pool", slog.String("pool", pool.Name))
```
gibmat: I feel like we should have identical logic in both trim functions. Maybe a helper function of this loop's body that can be called from both trim functions?

incus-osd/internal/zfs/zfs.go
```diff
@@ -1087,6 +1089,73 @@ func ScrubAllPools(ctx context.Context) error {
...
+	_, pools, err := storage.GetStorageInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	pool := api.SystemStoragePool{}
+
+	for _, p := range pools {
+		if p.Name == poolName {
+			pool = p
```
gibmat: Can have a `break` here as a small optimization.

incus-osd/internal/storage/storage.go
```diff
@@ -64,6 +64,42 @@ func zpoolScrubStateToPoolScrubState(state ZpoolScrubState) api.SystemStoragePoo
...
+	ZpoolTrimActive ZpoolTrimState = "ACTIVE"
+	// ZpoolTrimCanceled represents that the vdev trim was canceled.
+	ZpoolTrimCanceled ZpoolTrimState = "CANCELED"
+	// ZpoolTrimSuspended represents that the vdev trim was suspended.
+	ZpoolTrimSuspended ZpoolTrimState = "SUSPENDED"
+	// ZpoolTrimComplete represents that the vdev trim has finished.
+	ZpoolTrimComplete ZpoolTrimState = "COMPLETE"
+)
+
+var zpoolToTrimStateMap = map[ZpoolTrimState]api.SystemStoragePoolTrimState{
```
gibmat: Is there a specific reason why we're introducing two almost identical sets of constants for reporting trim status?

gibmat (changes requested):

incus-osd/internal/zfs/zfs.go
stgraber: yep

incus-osd/internal/storage/storage.go
stgraber: basically a 1:1 match to the scrub logic but I'll look at doing some cleanup

incus-osd/internal/zfs/zfs.go
stgraber: yeah, current pass was basically identical logic to keep things easy and consistent, but some helpers make sense

gibmat (approved):

PR #1328, opened 2026-08-28
Install / UI fixes

stgraber:


incus-osd/internal/tui/tui.go
```diff
@@ -434,11 +426,34 @@ func wrapFooterText(label string, text string, maxLineLength int) []string {
...
 
+// discoverConsoleDevices adds any additional console devices to the list of ttys used by the TUI.
+func discoverConsoleDevices(s *state.State) {
+	// If we're running in an Incus VM, additionally use /dev/ttyS0.
+	_, err := os.Stat("/dev/virtio-ports/org.linuxcontainers.incus")
+	if err == nil && !slices.Contains(ttyDevs, "/dev/ttyS0") {
+		ttyDevs = append(ttyDevs, "/dev/ttyS0")
+	}
+
+	if s == nil {
```
gibmat, line 438: It would be good to add a comment here about why we return early if IncusOS is running within an Incus VM. :)

gibmat (changes requested):

gibmat (approved):

PR #1329, opened 2026-08-29
issue 1300 correction. Add message before wipe operation

### LINKED ISSUE #1300: Add additional status message during installation to report disk wipe
It would be useful to show a different status message like "Wiping target driver, this may take a while" so that folks don't think we're stuck when dealing with a very slow blkdiscard operation.

MaksUnweb:
Closed: #1300


stgraber:
Closing as this has already been implemented in #1328

PR #1334, opened 2026-09-02
Improve Ceph handling

stgraber:


incus-osd/internal/ceph/ceph.go
```diff
@@ -1162,3 +1341,34 @@ func getOSDHosts(ctx context.Context) ([]string, error) {
...
+
+	count := 0
+
+	for _, i := range instances {
+		if !strings.HasPrefix(i.Name, "ceph-osd-") {
+			continue
+		}
+
+		for _, device := range i.Devices {
+			if device["type"] == "unix-block" {
```
gibmat: Shouldn't we also check that the device's name has the `ceph-` prefix too, so we don't count other random block devices? That shouldn't happen, but we also can't control the end-user adding things to the container(s), and it's an easy enough check to guard against accidentally allow wiping too many OSDs.

incus-osd/internal/ceph/ceph.go
```diff
@@ -420,30 +479,103 @@ func RemoveOSD(ctx context.Context) error {
...
+	// Remove a single drive if others remain on this host.
+	//
+
+	if deviceName != "" && len(localDrives) > 1 {
+		driveCount, err := getOSDDriveCount(ctx)
+		if err != nil {
+			return err
+		}
+
+		if driveCount <= 3 {
```
gibmat: We now have two different conditions where we return the same error, here (number of total OSD drives) and later (number of OSD instances). We should probably be consistent and have just a single check right before this block of code that can cover both removal of a drive from an OSD with multiple drives, or the removal of an OSD container itself.

incus-osd/internal/ceph/ceph-refresh.sh
```diff
@@ -1,8 +1,16 @@
 #!/bin/sh
 # shellcheck disable=SC1083
 
+# Setup the object gateway on containers predating its introduction
```
gibmat, line 4: Should probably annotate that this code can be removed after some period of time, since it won't be needed after any existing Ceph deployment has been refreshed.

doc/tutorials/storage-create-ceph-cluster.md
```diff
@@ -36,9 +36,9 @@ Ceph requires a minimum of three OSDs. Each Incus server runs at most one OSD in
 Each OSD instance is bound to a specific Incus server, so the `--target` parameter must be specified when adding a drive. Repeating the action with additional drives on the same server attaches them to the existing OSD instance.
 
 ```
-incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server01
-incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server02
-incus admin os application action incus-ceph -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server03
+incus admin os application action incus -d '{"action":"add-drive","config":{"device_id":"/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_incus_ceph--disk"}}' --target server01
```
gibmat: Hmm, seeing this now I realize it's pretty confusing -- I'm going to `add-drive` to Incus, yay! Wait, why am I getting a random error back about Ceph or OSD??

I think the action names that were moved to the Incus application should be modified to include a `ceph-` prefix or similar.

incus-osd/internal/ceph/ceph.go
```diff
@@ -1110,104 +1100,13 @@ func getInstanceIPv6Addr(ctx context.Context, instanceName string) (string, erro
...
 	incusClient, err := incus.ConnectIncusUnixWithContext(ctx, "", nil)
 	if err != nil {
 		return err
 	}
 
 	incusClient = incusClient.UseProject(projectName)
 
-	var existingImageFingerprint string
-
-	// Determine if the OCI image currently exists locally.
```
gibmat, line 957: I haven't checked the Incus code, but hopefully it's got optimized logic in place that first checks an OCI image's fingerprint and only downloads the potentially large image if the remote OCI server reports a different fingerprint? In my first implementation of this code I wasn't doing that, and forcefully downloading identical OCI images when refreshing Ceph containers was both wasteful of bandwidth and made the API call super slow.

gibmat: Hmm, maybe this is addressed in the next commit. 🙂

gibmat (changes requested):

incus-osd/internal/ceph/ceph.go
stgraber: Hmm, yeah, we can add an extra check. I'm sure the user would be in for a bad time eventually if they go mess with that stuff, but it's a cheap check to add :)

incus-osd/internal/ceph/ceph-refresh.sh
stgraber: Yeah, we can get rid of it in 3 months. Realistically we could almost certainly remove it already given I'm likely the only one using this logic, but let's be safe :)

doc/tutorials/storage-create-ceph-cluster.md
stgraber: Oops, yeah, agreed, we want a service prefix on those. Would get very confusing if/when we add Linstor too ;)

incus-osd/internal/ceph/ceph.go
stgraber: Yeah, Incus computes the fingerprint based on the OCI digest. It's actually a bit of a lie as the fingeprint it tracks actually differs from the actual fingerprint on disk (yeah, that's ugly) but that's done specifically so we can correctly check for changes.

stgraber:
@gibmat addressed the review comments. Will want to retry those two test failures, looks like it's coming from some Github 500s...

gibmat (approved):

PR #1339, opened 2026-09-02
Cleanup rootfs from initrd

stgraber:


stgraber:
@gibmat failure looks like it's triggered because of the incus-agent config, any idea why it would hit after the update?

gibmat:
> @gibmat failure looks like it's triggered because of the incus-agent config, any idea why it would hit after the update?

I'm not sure; it should behave as if incus-agent isn't configured and then automatically generate the necessary configuration file. Running `make && make test && make && make test-upgrade` locally will probably be the easiest way to see any errors that might be logged.

incus-osd/cmd/initrd-utils/cleanup.go
```diff
@@ -0,0 +1,141 @@
...
+	_, err := os.Lstat(deletedPath)
+	firstRun := err != nil
+
+	if firstRun {
+		err := os.Mkdir(deletedPath, 0o500)
+		if err != nil {
+			return err
+		}
+
+		err = os.Chmod(deletedPath, 0o500) //nolint:gosec
```
gibmat: Isn't this redundant?

stgraber:
@gibmat may just be racy?

gibmat (changes requested):

gibmat:
Well, something's causing the initrd to wedge. Can reproduce with a simple reboot after install, no need to perform an upgrade.

stgraber:
@gibmat That should take care of the issue by having the test wait until incus-osd is up and allows the access again.

I also noticed that the fuse file was missing in the allow list, so fixed that.

And then went a bit of a rabbit hole around nvme and iscsi identifiers. They were only partly stable currently, so we'd have reset them and caused a fair bit of breakage. Change our generation logic so that they will be stable on new installs at least, but we can really do anything to the existing ones without breaking users.

stgraber:
Other problem was that the systemd presets only apply if /etc/machine-id doesn't exist, so had to move all the presets to /usr/lib/systemd instead.

.github/workflows/tests.yml
```diff
@@ -208,6 +214,9 @@ jobs:
 
           sleep 3m
 
+          # Wait for incus-osd to re-enable the incus-agent features following the reboot.
```
gibmat, line 217: If we're switching to polling for incus-agent to be running, let's remove/replace the static `sleep 3m` statements above. Should help things to run a bit quicker. 😀

gibmat (approved):

.github/workflows/tests.yml
stgraber: Well, except we don't know how long it will take for the current incus-agent to stop responding. So we'd still need a sleep or another loop to figure out when the system is actually restarting, then do the loop.

PR #1344, opened 2026-09-04
incus-osd/network: Reject configured MAC addresses that don't exist on any known device

### LINKED ISSUE #1028: Setting a bad MAC address for network device breaks further network configuration updates until reboot
This is likely a pretty obscure edge case, but I noticed while testing that if I set an invalid MAC address for an existing network device, `networkctl` gets confused because it thinks the network device already exists. The journal log contains this warning:

> enp5s0: Failed to create netdev: File exists

The changes are properly rejected, but attempting to make another change via the API or relying on a rollback timeout fail with the error "timed out waiting for network to come online".

The root cause seems to be udev detecting the removed `_p<good MAC>.link` file, but then when `_p<bad MAC>.link` is generated defaulting to using `/usr/lib/systemd/network/99-default.link` and never properly being reconfigured with the good MAC, which then breaks our networking configuration.

gibmat:
Closes #1028

### DIFF
--- a/incus-osd/internal/network/networkd_validate.go
+++ b/incus-osd/internal/network/networkd_validate.go
@@ -1,6 +1,7 @@
 package network
 
 import (
+	"context"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -13,7 +14,7 @@ import (
 	"github.com/lxc/incus-os/incus-osd/api"
 )
 
-func validateInterfaces(interfaces []api.SystemNetworkInterface, requireValidMAC bool) error {
+func validateInterfaces(ctx context.Context, interfaces []api.SystemNetworkInterface, requireValidMAC bool) error {
 	for index, iface := range interfaces {
 		err := validateName(iface.Name)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


incus-osd/internal/network/networkd_validate.go
```diff
@@ -518,6 +543,26 @@ func validateHwaddr(hwaddr string, requireValidMAC bool) error {
...
 
+func hwaddrExists(ctx context.Context, hwaddr string) error {
+	// To support testing, if an empty context is provided, pretend the device exists.
+	if ctx == context.TODO() {
+		return nil
+	}
+
+	// `ip link show` will report all known network interfaces and their MACs.
+	// We can then do a simple search to see if the provided MAC is in the output.
+	output, err := subprocess.RunCommandContext(ctx, "ip", "link", "show")
```
stgraber: Any reason not to use `net.Interfaces()` and then match on `iface.HardwareAddr`?

stgraber (approved):

PR #1347, opened 2026-09-08
incus-osd/internal/state: Update zpool keylocation when moving key files

stgraber:


incus-osd/internal/state/upgrades.go
```diff
@@ -222,8 +226,34 @@ System.Network.Config.Proxy.Rules[%d].Target: direct
...
+	_, err := subprocess.RunCommandContext(ctx, "zpool", "import", pool)
+	if err != nil && !strings.Contains(err.Error(), "cannot import '"+pool+"': a pool with that name already exists") {
+		slog.Warn("Unable to import storage pool to update its key location", "pool", pool, "err", err)
+
+		return nil
+	}
+
+	_, err = subprocess.RunCommandContext(ctx, "zfs", "set", "keylocation=file:///var/lib/incus-os/keys/zpool."+pool+".key", pool)
+
+	return err
```
gibmat, line 258: We'll also need to then export the pool, otherwise later on during startup `zfs.LoadPools()` will error out when attempting to import a pool that's already imported.

gibmat (changes requested):

incus-osd/internal/state/upgrades.go
stgraber: @gibmat LoadPool has logic to skip on already imported

gibmat: Ah, missed that negation in the import check. :slightly_smiling_face:

gibmat (approved):

PR #1348, opened 2026-09-09
Add storage seed and allow configuring storage pool alignment

gibmat:
This introduces an new `storage` seed that allows for configuring the creation of additional storage pools on first boot, rather than needing to rely on creating these pools via API later on. Limited configuration of the special "local" pool is also allowed via this seed.

The storage API also has a new field `Alignment` that allows for the configuration of a new storage pool's sector alignment. Any valid power of two is allowed; it is assumed the user will properly understand the implications of setting an improper alignment value when creating their storage pool(s). If no alignment is configured, IncusOS will continue to default to an alignment value of 4KiB.

### DIFF
--- a/incus-osd/api/seed/storage.go
+++ b/incus-osd/api/seed/storage.go
@@ -1,0 +1,13 @@
+package seed
+
+import (
+	"github.com/lxc/incus-os/incus-osd/api"
+)
+
+// Storage represents the storage seed used to automatically
+// create storage pools on first boot.
+type Storage struct {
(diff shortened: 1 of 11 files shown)


incus-osd/api/seed/storage.go
```diff
@@ -0,0 +1,8 @@
+package seed
+
+// Storage represents the storage seed.
+type Storage struct {
```
stgraber, line 4: Is that the first seed where the seed struct differs from the API struct of the same name?

Couldn't we have the storage seed allow a more standard api.SystemStorage type data?

PR #1341, opened 2026-09-03
Extra hardening

stgraber:

PR #1351, opened 2026-09-09
base: Delay enabling kernel `lockdown=integrity` mode

gibmat:
On systems with SecureBoot enabled, the kernel will automatically be in `lockdown=integrity` mode and able to properly load signed kernel modules such as ZFS. But on systems without SecureBoot, setting this as a kernel command line option prevents the loading of the ZFS module. Instead, immediately after loading the zpools (which will trigger a loading of the module), write to the `/sys/` knob to enable integrity mode which cannot then be turned off without a reboot.

Related to #1341


stgraber (approved):

rapenne-s:
I confirm that removing `lockdown=integrity` from the kernel command line allows incus os to boot again on systems without secure boot.

PR #1364, opened 2026-09-13
incus-osd/network: support on-link network routes

### LINKED ISSUE #1359: network seed cannot configure on-link gateways for static /32 routes
## Context

I am trying to set up IncusOS on dedicated cloud hosts where I have no control over public-IP ranges, gateways, subnets, etc. 

## Network seed

```yaml
interfaces:
  - name: "eth0"
    hwaddr: "<VIRTUAL-MAC>"
    addresses:
      - "<STATIC-IP>/32"
    routes:
      - to: "0.0.0.0/0"
        via: "<OFF-SUBNET-GATEWAY>"
```

## Actual behavior

The static address is assigned successfully, but no default route is installed. All executed via `incus exec ...`

```console
$ ip -4 address show
...
inet <STATIC-IP>/32 ...

$ ip -4 route show
# No default route

$ ip route get 1.1.1.1
RTNETLINK answers: Network is unreachable
```

`networkctl` reports the interface as:

```text
State: routable (configuring)
```

The generated networkd configuration contains:

```ini
[Route]
Gateway=<OFF-SUBNET-GATEWAY>
Destination=0.0.0.0/0
```

For a `/32` address whose gateway is outside the address prefix, the route
requires:

```ini
[Route]
Gateway=<OFF-SUBNET-GATEWAY>
GatewayOnLink=yes
Destination=0.0.0.0/0
```

## Workaround

The route works when added manually:

```bash
ip route replace default \
  via <OFF-SUBNET-GATEWAY> \
  dev <INTERFACE> \
  onlink
```

After adding the route, connectivity works. However, the route is lost after network reconfiguration or reboot.

## Request

Please consider adding a field to [_SystemNetworkRoute_](https://github.com/lxc/incus-os/blob/d98d7e0a335197dbe1659f6cf21c087419e816eb/incus-osd/api/system_network.go#L161) to mark the gateway as on-link.

## Suggestions

`GatewayOnLink bool json:"gateway_onlink,omitempty"...` OR `OnLink bool json:"onlink,omitempty"...`

hangrymuppet:
Fixes https://github.com/lxc/incus-os/issues/1359

Allow off-prefix gateways through the network seed. Emit GatewayOnLink=yes in systemd-networkd configuration, preserve the on-link flag in route state, and reject invalid dynamic or missing gateways.


hangrymuppet:
Should `hwaddr` be a MAC address in the docs? It is currently [`enp5s0`](https://github.com/lxc/incus-os/pull/1364/changes#diff-c85eb59f3b21e898b6e8488932adf0d9780ce8bd267b76ce6e1271b89028024bR124) and I followed the pattern.

stgraber:
> Should `hwaddr` be a MAC address in the docs? It is currently [`enp5s0`](https://github.com/lxc/incus-os/pull/1364/changes#diff-c85eb59f3b21e898b6e8488932adf0d9780ce8bd267b76ce6e1271b89028024bR124) and I followed the pattern.

It must be a MAC address on running systems but can be an interface name in the seed data.

incus-osd/internal/network/networkd.go
```diff
@@ -521,20 +536,12 @@ func getWireguardState(ctx context.Context, iface string) (api.SystemNetworkInte
 	}
 
 	// Get routes for the interface.
-	routes := []api.SystemNetworkRoute{}
-	routeRegex := regexp.MustCompile(`(.+) via (.+) proto`)
-
 	output, err := subprocess.RunCommandContext(ctx, "ip", "route", "show", "dev", resolveBridge(iface))
```
gibmat, line 539: Let's move running of the command into the `parseRoutes()` method too, since nothing else uses the resulting output.

incus-osd/internal/network/networkd_validate.go
```diff
@@ -51,6 +51,10 @@ func validateInterfaces(ctx context.Context, interfaces []api.SystemNetworkInter
 		}
 
 		for routeIndex, route := range iface.Routes {
+			routeValidationErr := validateOnLinkRoute(route)
```
gibmat, line 54: We tend to use use `err` when checking for returned errors; not sure if the extra verbosity is worthwhile here and the next couple of changes.

gibmat (changes requested):
Can you please split your changes in to a few commits? (Adding the API field / validation logic/tests, updating documentation, and updating the generated rest-api.yaml file.)

incus-osd/internal/network/networkd.go
hangrymuppet: Done.

I've still kept a helper to do the regex parsing so that can be unit tested; my primitive brain has an ick whenever I have to resort to something like that instead of an api with a contract but then my pragmatic brain says that everything cannot be perfect always.

Let me know if you prefer I didn't do that and kill the unit test. Happy to do either :)

incus-osd/internal/network/networkd_validate.go
hangrymuppet: Fixed

hangrymuppet:
> Can you please split your changes in to a few commits? (Adding the API field / validation logic/tests, updating documentation, and updating the generated rest-api.yaml file.)

Split complete... (now only if I could do the other split 🤣 ). Leaving the other comments open; please resolve them if you are happy with the changes.

Pasted 2026-09-15 at 11 38 07

Stay cool!

hangrymuppet:
Whoops; sorry. Forgot about the last CI run that also flagged static analysis issues. Fixed with new commit.

incus-osd/internal/network/networkd.go
gibmat: I think I'd prefer to just have a single function `parseRoutes()` that's self-contained at the expense of a unit test. We don't really have many (any?) unit tests that attempt to directly check the output of commands, so trying to include one here feels a bit out of place.

incus-osd/internal/network/networkd.go
```diff
@@ -522,10 +522,11 @@ func parseRoutes(ctx context.Context, iface string) ([]api.SystemNetworkRoute, e
 }
 
 func parseRouteCommandOutput(output string) []api.SystemNetworkRoute {
-	routes := []api.SystemNetworkRoute{}
```
gibmat: Please squash-merge these changes into the main commit -- keeps the commit history cleaner. :)

incus-osd/internal/network/networkd.go
hangrymuppet: Done.

> We don't really have many (any?) unit tests that attempt to directly check the output of commands

I urge you to reconsider in the future for dependencies that rely on parsing raw text without an api contract.

incus-osd/internal/network/networkd.go
hangrymuppet: I was being lazy :D good man!

gibmat (approved):

Issue #988, opened 2026-03-25
Add support for multi-profile UKI


stgraber:
UKI support multiple profiles which we could use to provide a number of different kernel command line variations.

https://github.com/systemd/particleos/tree/main/mkosi.uki-profiles

One thing that came up in the past through the forum was IOMMU tweaks for getting virtual IOMMU groups on consumer grade devices allowing GPU passthrough in environments that wouldn't otherwise support it.

gibmat:
Once #1384 lands, we'll have the foundation needed to start using UKI profiles. If we come up with a list of various kernel command lines we'd like to add support for (such as the IOMMU tweaks mentioned above) I can prepare those profiles and use that PR to close this issue. It's also possible to trigger additional logic from `incus-osd` based on the UKI profile, so we're not limited to just kernel command line flags.

PR #1384, opened 2026-09-21
Preliminary UKI profile support

gibmat:
This adds preliminary support for IncusOS to begin using UKI profiles. While no behavior will change until a build actually defines a profile, logic has been added to allow existing IncusOS systems to seamlessly transition to the default UKI profile ("main") without any intervention.

IncusOS will attempt to automatically (re)boot into the same profile that it's currently running, as well as reporting the UKI profile name in the initial startup message.

Partly addresses #988


incus-osd/internal/secureboot/pcrs.go
```diff
@@ -170,7 +170,7 @@ func ForceUpdatePCRBindings(ctx context.Context, osName string, osVersion string
 	}
 
 	// Once complete, immediately reboot the system which should then auto-unlock.
-	_, err = subprocess.RunCommandContext(ctx, "systemctl", "reboot")
+	err = util.SystemReboot(ctx)
```
stgraber, line 173: The PCR rebind can happen late, after applications and services are running, it should instead trigger the channel to ensure the correct shutdown order.

gibmat:
Fun... a new linter release...

stgraber:
> Fun... a new linter release...

We've seen worse :)

